PR #19, opened 2015-06-30
Purl interface will render webarchive-seed as image type

aalsum:


spec/model/purl_object_spec.rb
```diff
@@ -5,4 +5,25 @@
     item = PurlObject.new('bb157hs6068')
     expect(item).to be_ready
   end
+  
+  describe ".image?" do
+    it "should return true for the content type webarchive-seed" do
+      item = PurlObject.new('ab123cd4567')
+      item.instance_variable_set(:@type,"Image")
```
jkeck: It doesn't appear that this test is actually testing what it says it is.

The `it` block says that it should return true for "webarchive-seed" type, however in the test the `@type` ivar is being set to "Image"  Perhaps there should be an additional test that sets the `@type` ivar to "webarchive-seed" and validate that the item responds appropriately to `.image?`.

spec/model/purl_object_spec.rb
```diff
@@ -5,4 +5,25 @@
...
+      item.instance_variable_set(:@extracted,true)
+      expect(item.image?).to be_truthy
```
jkeck: Very minor comment but you should also be able to write this as `expect(item).to be_image` since the object responds to `image?`

PR #20, opened 2015-07-16
Fix production

cbeer:


jkeck:
I don't quite understand d46f8d31f58ada626a638a0deb3b4de97e1e5a76 (`with_format` in particular) but I don't quite understand why it was like that in the first place, and it is certainly more understandable than it was.

PR #41, opened 2015-11-02
Stop generating flipbook metadata

cbeer:


cbeer:
Before merging this, we need to remove any other flipbook consumers.

Issue #45, opened 2015-11-04
Selected URI Pattern for IIIF Annotations


blalbrit:
pattern should be: {druid}/iiif/canvas-n/annotations per conversation with @azaroth42 and @LynnMcRae 

This is a back-burner for now.

azaroth42:
To clarify...

The URI pattern for Annotation __Lists_ (see http://iiif.io/api/presentation/2.0/#other-content-resources) could be `{druid}/iiif/canvas-{n}/annotations`.  The annotations in that list should be those that have `canvas-{n}` or part of it as their target.  However this limits implementation possibilities as canvas-{n} cannot be both a directory and a file (the JSON) at the same time.

The URI of the Annotations in that list has not been discussed, but I would propose one of:

  * {druid}/iiif/canvas-{n}/anno-{m}
  * {druid}/iiif/canvas-{n}/annotations/anno-{m}

My concern with the second is the same as with the current proposed pattern for lists.

cbeer:
IIIF now recommends some URI patterns: http://iiif.io/api/presentation/2.1/#a-summary-of-recommended-uri-patterns

I suggested we move from our current IIIF annotation URI pattern of using sequential numbering for our canvas URIs to a more permanent URI structure using the resource id already present in the contentMetadata.

PR #46, opened 2015-11-04
Use selected IIIF canvas and annotation urls; fixes #44 and #45

### LINKED ISSUE #44: Selected URI Pattern for IIIF Canvases
pattern should be: {druid}/iiif/canvas-{n} , where {n} starts with 1. per conversation with @azaroth42 and @LynnMcRae 

Should dereference the json fragment of the canvas as it appears within the manifest.json

cbeer:


azaroth42:
Was the PR reviewed?

Issue #49, opened 2015-12-02
add support for externalFile references for IIIF manifest generation


drh-stanford:
This is for the virtual objects (atlases). See the old PR here: sul-dlss/dor-services#89

The specification is here: https://consul.stanford.edu/display/chimera/Support+for+virtual+resources

Here is the prototype object in production: https://argo.stanford.edu/catalog/druid:hj097bm8879

Took a cursory look at this issue and have the following list of tasks:
- [x] establish development environment on laptop
- [x] add fixture data for virtual parent object and children
- [x] add unit tests in `iiif_manifest_spec.rb` for virtual objects with only remote resources
- [x] change `ContentMetadata.extract_resources` to handle child objects from `externalFile` data
- [ ] change/verify dor-services (publish step) exports correct metadata for externalFiles (see https://github.com/sul-dlss/dor-services/pull/128)
- [x] PR review/merge
- [ ] migrate prod Rumsey virtual atlas to stage environment
- [ ] deploy to stage
- [ ] deploy to prod

drh-stanford:
Took a cursory look at this issue and have the following list of tasks:

- [ ] establish development environment on laptop
- [ ] change `ContentMetadata.extract_resources` to handle child objects from `externalFile` element data
- [ ] add fixture data for virtual parent object and children
- [ ] add unit tests in `iiif_manifest_spec.rb` for virtual objects without file resources, and with mix of file and externalFile
- [ ] PR review/merge
- [ ] migrate prod Rumsey virtual atlas to stage environment
- [ ] deploy to stage
- [ ] deploy to prod

cbeer:
I assume the outstanding tasks have all been done by now?

drh-stanford:
yes, closing

PR #51, opened 2015-12-08
implementation of remote resources in IIIF manifests for virtual objects

drh-stanford:
This PR implements virtual object support for IIIF manifests. See issue #49 for details.

@cbeer @tingulfsen I'd like to pair up to review this PR as my testing was limited to my laptop and I did not deploy to the -dev or -test environments.

Note the Travis build is failing whereas my local instance passes all specs. I did run into configuration issues as my virtual object fixture data is not in the -prod or -test environment (which may be why Travis is failing if it's trying to fetch public XML data from outside the fixtures directory) but I was able to change my local configuration to work solely off my laptop for testing it. I've added 3 new feature tests for 2 child objects and the parent virtual object.

```
IIIF manifests
  works
  includes a representative thumbnail
  includes authorization services for a Stanford-only image
  suppresses sequences for dark resources
  virtual objects
    works for the first child object
    works for the second child object
    works for the parent object
```

### DIFF
--- a/app/models/content_metadata.rb
+++ b/app/models/content_metadata.rb
@@ -21,6 +21,17 @@ class ContentMetadata
     document.at_xpath('bookData') if document
   end
 
+  ##
+  # The content metadata has a druid for the resources which in practice
+  # is the same for the overall item, but with virtual objects, the druid
+  # for the resources changes to child items. So we include the druid at
+  # the content metadata Resource-level to accommodate it.
+  #
+  # @return [String]
+  def druid
+    document.at_xpath('@objectId').to_s.gsub(/^druid:/, '') # required attribute
+  end
+
   def resources
     return [] unless document
 
@@ -37,14 +48,23 @@ class ContentMetadata
     resources.select { |r| r.type == 'object' }
   end
 
+  # rubocop:disable Metrics/AbcSize
   def extract_resources(resource)
+    # extract resource-level attributes first
     resource_attributes = {
+      id: resource.attribute('id').to_s,
       type: resource.attribute('type').to_s,
-      label: resource.xpath('(label|attr[@name="label"])/text()').first.to_s
+      label: resource.xpath('(label|attr[@name="label"])/text()').first.to_s,
+      druid: druid
     }
 
     files = resource.xpath('file').select { |file| Purl::Util.file_ready? file }.map do |file|
-      Resource.from_content_metadata(file, resource_attributes)
+      Resource.from_file_metadata(file, resource_attributes)
+    end
+
+    # External files point to "remote" objects that themselves hold the resource
+    external_files = resource.xpath('externalFile').map do |external_file|
+      Resource.from_external_file_metadata(external_file, resource_attributes)
     end
 
     resources = resource.xpath('resource').map do |nested_resource|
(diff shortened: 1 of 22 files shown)


app/models/content_metadata.rb
```diff
@@ -74,6 +89,29 @@ def self.from_content_metadata(file, options = {})
...
+    # Extract resource attributes from `<externalFile>...</externalFile>` in content metadata
+    #
+    # @example
+    #   <externalFile fileId="2542A.jp2" objectId="druid:cg767mn6478" resourceId="cg767mn6478_1" mimetype="image/jp2">
+    #     <imageData width="6475" height="4747"/>
+    #   </externalFile>
+    #
+    def self.from_external_file_metadata(externalFile, options = {})
+      new(
+        options.merge(
```
cbeer: There are suspicious similarities with `.from_file_metadata`. Is there some way we could share the logic between the two implementations?

app/models/iiif_presentation_manifest.rb
```diff
@@ -98,8 +98,10 @@ def body(controller = nil)
     manifest
   end
 
+  ##
+  # @param [Integer] count This is the i_th (image) resource
   def canvas_for_resource(purl_base_uri, resource, count)
-    url = stacks_iiif_base_url(druid, resource.filename)
+    url = stacks_iiif_base_url(resource.remote_druid.present? ? resource.remote_druid : druid, resource.filename)
```
cbeer: Could we push this logic upstream into `Resource`? Maybe as `Resource#druid`, and it just does the right thing?

cbeer:
I'm also getting the same test failure that travis is. Is it possible there's some file or code that wasn't pushed?

drh-stanford:
Re: travis errors. Have you seen this `Couldn't find template for digesting: purl/manifest` error(?) message before? I don't have a `manifest` file checked in for my fixtures, but I'm assuming dynamic generation of it. The `bb157hs6068` druid has a `manifest` file in its fixtures but it also gets the same message.

```
Started GET "/bb157hs6068/iiif/manifest.json" for ::1 at 2015-12-09 11:56:46 -0800
Processing by PurlController#manifest as HTML
  Parameters: {"id"=>"bb157hs6068"}
  Couldn't find template for digesting: purl/manifest
Completed 200 OK in 9ms (Views: 0.7ms | ActiveRecord: 0.0ms)

Processing by PurlController#manifest as HTML
  Parameters: {"id"=>"hj097bm8879"}
  Couldn't find template for digesting: purl/manifest
  Rendered text template (0.0ms)
Completed 404 Not Found in 8ms (Views: 0.3ms | ActiveRecord: 0.0ms)

Started GET "/jw923xn5254/iiif/manifest.json" for ::1 at 2015-12-09 11:54:30 -0800
Processing by PurlController#manifest as HTML
  Parameters: {"id"=>"jw923xn5254"}
  Couldn't find template for digesting: purl/manifest
Completed 200 OK in 8ms (Views: 0.8ms | ActiveRecord: 0.0ms)
```

drh-stanford:
Re: travis errors. the problem is that `hj097bm8879` isn't on production purl, only in the fixtures directory. so the route yields a 404 since it tries to fetch the public xml, etc. from production purl.

drh-stanford:
I've tried to address your comments in this updated PR. I fixes the Travis errors, DRY'd up some code, and migrated the druid value in the IIIF manifest to content metadata as a delegate to accomodate the changing druid in externalFiles. It's a fairly big PR, so please let me know if you have questions/comments or want to pair up.

app/models/content_metadata.rb
```diff
@@ -82,5 +126,20 @@ def levels
       return unless width > 0 && height > 0
       ((Math.log([width, height].max) / Math.log(2)) - (Math.log(96) / Math.log(2))).ceil + 1
     end
+
+    private_class_method
```
cbeer: I haven't seen this pattern before -- does it actually work, or, do you know what rubies it works in? My naive testing in ruby 2.2.3 doesn't seem to do anything:

```ruby
class A
private_class_method
def self.a; 1; end
end

A.a # => 1
```

app/models/content_metadata.rb
```diff
@@ -1,5 +1,5 @@
 class ContentMetadata
-  attr_reader :document
+  attr_reader :document, :druid
```
jkeck: I don't believe we need to explicitly add `druid` as an `attr_reader` if we have defined the method below.  I think that you'll find the behavior is functionally equivalent without this.

Typically the use case for `attr_reader` would be if you were setting `@druid` in the `initialize` method and wanted to be able to access the value by `druid` instead of `@druid` (as well as, in this case, be available to be called on instances of the class outside of the class).

app/models/content_metadata.rb
drh-stanford: Looks like you have to break it out, like so
```
[5] pry(main)> class C
[5] pry(main)*   def self.c; 'hello'; end
[5] pry(main)*   private_class_method :c
[5] pry(main)* end  
C < Object
[6] pry(main)> C.c
NoMethodError: private method `c' called for C:Class
```

drh-stanford:
Ok, thanks for the comments. I tried to address that last two and I squashed and pushed up the latest branch.

PR #55, opened 2016-01-29
Remove primary/secondary mods name helpers and loop through all in…

jkeck:
… contributors section.

This is causing certain contributors to not be displayed (see "Morin, Andre" in jw402xz0487).

SearchWorks has unified all contributors into a re-labeled Creators/Contributors section.


cbeer:
:+1:

PR #64, opened 2016-07-29
add checks to is_it_working

ndushay:

PR #67, opened 2016-08-02
add checks to is_it_working

jmartin-sul:
a hopefully fixed version of PR #64 that gets the tests working again (64 should be superseded if this gets merged).


.rspec
```diff
@@ -0,0 +1 @@
+--colour
```
jkeck: Was this checked in intentionally w/ this PR?

config/initializers/is_it_working.rb
```diff
@@ -3,4 +3,10 @@
 
   # Check the memcache servers used by Rails.cache if using the DalliStore implementation
   h.check :dalli, cache: Rails.cache if defined?(ActiveSupport::Cache::DalliStore) && Rails.cache.is_a?(ActiveSupport::Cache::DalliStore)
+
+  h.check :url, get: Settings.stacks.url
+  h.check :url, get: Settings.stacks.iiif_profile
+
+  h.check :url, get: Settings.flipbook.url + '/' # response is 404 if trailing '/' is omitted
+  h.check :url, get: Settings.embed.healthcheck_url
```
jkeck: What would the effect of Embed's `is_it_working` check failing in this case?  Embed isn't a hard dependency of PURL (meaning PURL still technically works if Embed is down, it is just missing that feature).

Would this failure cause the load balancer to think PURL is down as well? (Pardon my `is_it_working` naiveté)

jmartin-sul: ah, good question... my guess is that it'd just result in a nagios alert, but to be totally honest, i'm not sure what-all pays attention to the result of `is_it_working` (and i was just trying to get the check that @ndushay added to work, didn't really consider whether all the checks were appropriate for `is_it_working`).

clarification from @eefahy maybe on the load balancer part?

.rspec
jmartin-sul: yup.  i split it out into a separate/secondary commit, but can split it out into a separate PR if that'd be preferable.  but i was running tests on my laptop a bunch while working on this, and color was a nice thing to add.

config/initializers/is_it_working.rb
jmartin-sul: on that note, should we be checking the flipbook URL?  what does flipbook do, exactly?

jkeck: I believe flipbook was used pre-embed as a book/image viewer.  I think it's still used in some legacy contexts though, and until we can migrate them all to embed we have to maintain it.

It is kind-of in the same boat I guess (although we have `is_it_working` checks for embed itself (that will trigger Nagios alerts, and I'm not sure about flipbook).

.rspec
jkeck: :shrug: I'm fine w/ it.  Just seemed odd to me.

config/initializers/is_it_working.rb
eefahy: Nagios is monitoring the `is_it_working` endpoints and expecting a 200 (for both purl and embed) The load balancer is checking `is_it_working` for a 200 as well as the string "Host:" (also for both purl and embed) Therefore, the content of the `is_it_working` page, other than the Host: string, has no effect on either the load balancer or the nagios checks.

jmartin-sul:
the `is_it_working` page returns a 500 if any of the checks fail, which would result in the webserver with the failing check getting pulled from rotation by the load balancer.  discussed w/ @jkeck and @cbeer on thurs and fri, and the consensus was that none of the services added to the check by this PR were hard enough dependencies that a node where that check is failing should be pulled.  and i didn't see a way to add an `is_it_working` check that would result in a status other than OK for the check without also resulting in a 500 for the page (`ok` and `fail` seem to be the only available statuses for a check).

so, it seems to me like this PR (and the one it builds on, PR #64) should be closed, since it seems like the new behavior that'd result isn't actually what we want.

PR #69, opened 2016-08-15
add code coverage services

ndushay:
* reorg Gemfile a bit
* actually send results to coveralls
* add CodeClimate and its test coverage also
* add github tagged release version badge


jmartin-sul:
FWIW:  i got a 404 on the code climate coverage page.  but it looks like the right link, because it's the same as the link from the main code climate purl page, and their internal link 404s also.

jmartin-sul:
and the aforementioned 404 is now a 200.  so all good on that front.

PR #75, opened 2016-08-18
update is_it_working with dependencies, including non-crucial ones

### LINKED ISSUE #73: purl: is_it_working tweaks
currently it is likely to cause problems in production due to my uninformed PR #624

- want it to have status 200 for load balancer or fail for hard dependencies only
- want "soft" dependencies (like was-thumbnail-service) shown in nagios.

Then update nagios checks, too.

Something like:

```
h.check :optional_server do |status|
  optional_status = IsItWorking::Status.new
  IsItWorking::UrlCheck.new(get: 'http://some.example.com/service').call(optional_status)
  optional_status.messages.each { |x| status.ok "#{'FAIL: ' unless x.ok?}#{x.message}" }
end
```

ndushay:
Also required updating searchworks url to https so is_it_working check would pass.

closes #73 
closes sul-dlss/media#112

@jmartin-sul @jkeck

### DIFF
--- a/config/initializers/is_it_working.rb
+++ b/config/initializers/is_it_working.rb
@@ -34,9 +34,12 @@ Rails.configuration.middleware.use(IsItWorking::Handler) do |h|
     non_crucial_url_check(Settings.embed.iframe.url_template.sub('{?url*}', "?url=#{resource_url}"), status, 'OEmbed service')
   end
 
-  h.check :non_crucial do |status|
-    non_crucial_url_check("#{Settings.searchworks.url}#{TEST_DRUID}", status, 'For "View in SearchWorks" link')
-  end
+  # FIXME: commenting this out for now.  Concerned about hammering SearchWorks with pings,
+  #   as these will all hit the Solr index.  This is minor functionality for purl.
+  #   When SW has a more benign endpoint for up-ness checks, we should use that.
+  # h.check :non_crucial do |status|
+  #   non_crucial_url_check("#{Settings.searchworks.url}#{TEST_DRUID}", status, 'For "View in SearchWorks" link')
+  # end
 
   h.check :non_crucial do |status|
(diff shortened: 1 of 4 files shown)


config/initializers/is_it_working.rb
```diff
@@ -1,6 +1,68 @@
...
+    non_crucial_url_check(Settings.stacks.url, status, 'Object content accessed here')
+  end
+
+  h.check :non_crucial do |status|
+    resource_url = Settings.embed.url.sub('%{druid}', TEST_DRUID)
+    non_crucial_url_check(Settings.embed.iframe.url_template.sub('{?url*}', "?url=#{resource_url}"), status, 'OEmbed service')
+  end
+
+  h.check :non_crucial do |status|
+    non_crucial_url_check("#{Settings.searchworks.url}#{TEST_DRUID}", status, 'For "View in SearchWorks" link')
```
jkeck: Perhaps this is a naive question, but I'm curious what the use for a SearchWorks check is on the PURL it-it-working page.  Looking over the code, it looks like there the only place the `Settings.searchworks.url` is used is a link when an object in PURL has a ckey.  So is this considered a soft dependency?  Is it useful for the service manager of PURL to see that SearchWorks is down from PURLs perspective (that they would not already be aware of otherwise)?

Just a few thoughts I had when reading over this.  I still may not have a full understanding on all the use-cases for it it working.

ndushay: There is a "View in SearchWorks" link;  the code to get there is diff if there is a ckey vs. when there isn't a ckey.  I figured one check would suffice for both;  and in both cases, the result is a "View in SearchWorks" link.  

I believe the concept is for a service manager to be aware that "view in searchworks" links won't work if searchworks is down.

PR #80, opened 2016-08-29
Replace multiple is_it_working checks on the local document cache…

jkeck:
…with comments about them being checked higher up.

### DIFF
--- a/config/initializers/is_it_working.rb
+++ b/config/initializers/is_it_working.rb
@@ -6,22 +6,11 @@ Rails.configuration.middleware.use(IsItWorking::Handler) do |h|
   # Check the memcache servers used by Rails.cache if using the DalliStore implementation
   h.check :dalli, cache: Rails.cache if defined?(ActiveSupport::Cache::DalliStore) && Rails.cache.is_a?(ActiveSupport::Cache::DalliStore)
 
-  # seems self-referential, but tells us if public_xml is working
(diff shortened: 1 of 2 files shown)


config/initializers/is_it_working.rb
```diff
@@ -6,22 +6,12 @@
   # Check the memcache servers used by Rails.cache if using the DalliStore implementation
   h.check :dalli, cache: Rails.cache if defined?(ActiveSupport::Cache::DalliStore) && Rails.cache.is_a?(ActiveSupport::Cache::DalliStore)
 
-  # seems self-referential, but tells us if public_xml is working
-  public_xml_url = Settings.purl_resource.public_xml.sub('%{druid}', TEST_DRUID)
-  public_xml_url = Settings.purl_resource.public_xml.sub('%{druid_tree}', TEST_DRUID) if public_xml_url.match(/druid/)
-  h.check :url, get: public_xml_url
+  # NOTE: Settings.purl_resource.public_xml exists in the document cache root on deployed enviornments, and that check should be reasonable
```
ndushay: "and the document cache root check should be sufficient" ?   instead of "that check should be reasonable" ?

ndushay:
This looks good;  I have a suggested wording tweak.

jkeck:
👍 I replaced the wording in the comment (and collapsed them down into one so I'm not repeating myself 3 times).

PR #85, opened 2016-09-13
README: ask for github release before deploying

ndushay:


jmartin-sul:
👍 (though also:  maybe this is better as a general guideline in the developer playbook?  seems like this would be a reasonable default suggestion for most apps, right?)

ndushay:
I put it in the readme because this app has NEVER had a github release, and in fact only has a single tag, which does not follow semantic versioning (`v3.0`).  Ugh.

jmartin-sul:
ah, yeah, fair enough.  will merge.  still might make for a good general admonishment.

PR #86, opened 2016-09-13
replace Squash with Honeybadger for exception notification

ndushay:
Note:  I already put honeybadger.yml files on the purl VMs so when this is deployed, it should work.


jkeck:
LGTM 👍 

Does anybody actively look at the Squash errors for PURL that we need to inform about this change before we :shipit:?

Issue #87, opened 2016-09-13
okcomputer to replace is_it_working


ndushay:
Exemplar:  purl-fetcher

possibly:
- `/status` endpoint for load-balancer needs and nagios errors
- `/status/[some-dependency]` for nagios warnings;  nagios routes them to approp parties
- `/status/all-dependencies` for full dependency documentation (?)

after this
- [x] purl PR merged to master
- [x] deploy
- [x] update Load Balancer (SAR-3250)
- [ ] update nagios 
  - [x] basic checks (~~sul-dlss/puppet#1001~~, ~~sul-dlss/puppet#1002~~, sul-dlss/puppet#1003, ~~sul-dlss/puppet#1018~~)
  - [x] purl-lb (sul-dlss/puppet/pull/1024)
  - [ ] any individual checks need separate nagios handling? (stacks, embed, feedback?) (#76)
- [x] remove is_it_working (#100)
  - [x] deploy
- [ ] update DevOpsDocs (now new nagios, loadbalancer ...) (sul-dlss/DevOpsDocs/pull/60)

ndushay:
I'm gonna call this DONE.

PR #90, opened 2016-09-14
okcomputer now mounted at /status

ndushay:
note that for seamless integration without coordinating deploy with load balancer and nagios updates, the code will have both is_it_working and okcomputer temporarily.

You can see this in action:

https://sul-purl-dev.stanford.edu/status
https://sul-purl-dev.stanford.edu/status/all
https://sul-purl-dev.stanford.edu/status/feedback_mailer   (or name of any specific check)

you can put `.json` at the end to get the results in json.

The plan:

- deploy this (after merge) so we can change load balancer and nagios to look for /status instead of /is_it_working
- after LB and nagios aren't dependent on is_it_working, remove it.

connects to #87 
connects to sul-dlss/media#125

### DIFF
--- a/spec/integration/purl_spec.rb
+++ b/spec/integration/purl_spec.rb
@@ -118,4 +118,22 @@ describe 'purl', type: :feature do
       expect(page).to have_text('OK')
     end
   end
+
+  describe '/status' do
+    it 'has response code 200' do
+      visit '/status'
+      expect(page.status_code).to eq 200
+      expect(page).to have_text('Application is running')
+    end
+    it '/status/all has response code 200' do
+      visit '/status/all'
+      expect(page.status_code).to eq 200
+      expect(page).to have_text('HTTP check successful')
+    end
+    it '/status/feedback_mailer responds' do
+      visit '/status/feedback_mailer'
+      expect(page.status_code).to eq 200
+      expect(page).to have_text('FeedbackMailer')
+    end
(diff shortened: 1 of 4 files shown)


jkeck:
👍 LGTM

spec/integration/purl_spec.rb
```diff
@@ -118,4 +118,22 @@
...
+    end
+    it '/status/all has response code 200' do
+      visit '/status/all'
+      expect(page.status_code).to eq 200
+      expect(page).to have_text('HTTP check successful')
+    end
+    it '/status/feedback_mailer responds' do
+      visit '/status/feedback_mailer'
+      expect(page.status_code).to eq 200
+      expect(page).to have_text('FeedbackMailer')
```
tingulfsen: In the other three tests, you check for some kind of success in the returned string, but not in this one?

tingulfsen: Except for the minor test question, it looks good to me.

tingulfsen:
Except for the minor test question, it looks good to me.

spec/integration/purl_spec.rb
ndushay: b/c I expect this test to fail - mailer won't be configured for testing, generally.  This way, the test is showing it's specific to FeedbackMailer whether the okcomputer check passes or fails.

PR #95, opened 2016-09-20
Update to Rails 5

cbeer:


config/environments/production.rb
```diff
@@ -22,7 +22,7 @@
 
   # Disable serving static files from the `/public` folder by default since
   # Apache or NGINX already handles this.
-  config.serve_static_files = ENV['RAILS_SERVE_STATIC_FILES'].present?
+  config.public_file_server.enabled = ENV['RAILS_public_file_server.enabled'].present?
```
jkeck: Was this done w/ an auto-generator or is there really a `RAILS_public_file_server.enabled` environment var?

ndushay: what was the answer to this question???

cbeer: yes, it was a mistake, and was corrected.

jkeck: Yeah, I'm not sure when GH stopped hiding comments on code that had been updated, but it's kinda confusing.  You have to look at the `Files changed` tab AFAICT.

PR #97, opened 2016-09-21
Allow druids on landing page to be configured per-env so production …

jkeck:
…druids don't cause 404 on non-production envs.

Currently sul-purl-dev and sul-purl-test landing pages return a 404 because it is attempting to load a production druid and the `PurlResource::ObjectNotReady` error is raised.

This allows us to simply set the druid list to an empty array (or nil) to remove the section entirely, or we can add environment specific druids to be displayed on the landing page for that environment.

## Before
before

## After
after

### DIFF
--- a/app/views/purl/index.html.erb
+++ b/app/views/purl/index.html.erb
@@ -25,40 +25,15 @@
   </div>
 </div>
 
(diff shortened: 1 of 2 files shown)


cbeer:
👏

PR #102, opened 2016-09-23
remove 'Available download formats' as it is now part of embed

### LINKED ISSUE #61: Drop "Available download formats" from display since it is now in Embed
See https://purl.stanford.edu/bb846gv7194. The "Available download file" appears to be redundant, and perhaps an artifact from pre-embed days where PURL was parsing and formatting content metadata.
screen shot 2016-04-25 at 1 45 40 pm

ndushay:
closes #61.  Section is redundant now that embed provides it.

## Before (https://purl.stanford.edu/bb846gv7194)

screen shot 2016-04-25 at 1 45 40 pm

## After

bb846gv7194 after removing download formats section


ndushay:
Note that coverage decreased because this PR removes some covered code from the overall code base.  I only removed the single test that checked for the presence of the now removed code.

jkeck:
This has been deployed to all purl environments except for production.

ndushay:
I am @ndushay and I approve @jkeck's previous message.

PR #112, opened 2017-01-06
Make IIIF canvases and annotations resolvable over HTTP using a more permanent url scheme

cbeer:


jkeck:
Minor comment, may be me just missing something.

app/models/content_metadata.rb
```diff
@@ -37,7 +37,7 @@ def resources
 
     @resources ||= document.xpath('resource').map do |resource|
       extract_resources(resource)
-    end.flatten.compact
+    end.flatten.compact.sort_by(&:sequence)
```
cbeer: Current behavior expects the resources to be sorted appropriately in the document, so this should actually present no functional change for our existing data.

PR #120, opened 2017-03-01
add all DC metadata in key/value pairs to IIIF manifests

### LINKED ISSUE #119: IIIF Manifests for Items: Add all DC fields as name/value pairs
Remove hardcoded name/value pairs for DC and just add them all.  Also switch publishVersion to the publish date.

peetucket:
this resolves #119

### DIFF
--- a/app/models/iiif_presentation_manifest.rb
+++ b/app/models/iiif_presentation_manifest.rb
@@ -169,10 +169,12 @@ class IiifPresentationManifest
 
   # transform all DC metadata in the public XML into an array of hashes for inclusion in the IIIF manifest
   def dc_to_iiif_metadata
-    all_dc_nodes = public_xml_document.xpath '//oai_dc:dc/*', 'oai_dc' => OAI_DC_SCHEMA
-    metadata = all_dc_nodes.map { |dc_node| iiif_key_value(dc_node.name.upcase_first, dc_node.text) }
-    metadata += public_xml_document.xpath('/publicObject/@published').map { |node| iiif_key_value('PublishDate', node.text) } # add published date
-    metadata
+    @dc_to_iiif_metadata ||= begin
+      all_dc_nodes = public_xml_document.xpath '//oai_dc:dc/*', 'oai_dc' => OAI_DC_SCHEMA
+      metadata = all_dc_nodes.map { |dc_node| iiif_key_value(dc_node.name.upcase_first, dc_node.text) }
+      metadata += public_xml_document.xpath('/publicObject/@published').map { |node| iiif_key_value('PublishDate', node.text) } # add published date
+      metadata
+    end
   end
 
   def iiif_key_value(label, value)
--- a/app/models/iiif_presentation_manifest.rb
+++ b/app/models/iiif_presentation_manifest.rb
@@ -6,6 +6,8 @@ class IiifPresentationManifest
 
   attr_reader :purl_resource
 
+  OAI_DC_SCHEMA = 'http://www.openarchives.org/OAI/2.0/oai_dc/'.freeze
+
   def initialize(purl_resource)
     @purl_resource = purl_resource
   end
@@ -69,15 +71,7 @@ class IiifPresentationManifest
     # Set viewingHint to paged if this is a book
     manifest.viewingHint = 'paged' if type == 'book'
 
-    metadata = []
-    # make into method, pass in xpath and label
-    metadata += get_metadata 'Creator', '//oai_dc:dc/dc:creator'
-    metadata += get_metadata 'Contributor', '//oai_dc:dc/dc:contributor'
-    metadata += get_metadata 'Publisher', '//oai_dc:dc/dc:publisher'
-    metadata += get_metadata 'Date', '//oai_dc:dc/dc:date'
-    metadata += get_metadata 'PublishVersion', '/publicObject/@publishVersion'
-
-    manifest.metadata = metadata if metadata.present?
+    manifest.metadata = dc_to_iiif_metadata if dc_to_iiif_metadata.present?
 
     manifest.description = description_or_note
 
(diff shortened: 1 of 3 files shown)


.rubocop_todo.yml
```diff
@@ -18,12 +18,18 @@ Lint/UselessAccessModifier:
...
+# Offense count: 1
+# Configuration parameters: CountComments.
+Metrics/BlockLength:
+  Max: 57
+
+# Offense count: 6
+# Configuration parameters: AllowHeredoc, AllowURI, URISchemes, IgnoreCopDirectives, IgnoredPatterns.
+# URISchemes: http, https
+Metrics/LineLength:
+  Max: 1046
```
cbeer: Instead of globally allowing 1000 characters per line, could you try excluding that single file?

app/models/iiif_presentation_manifest.rb
```diff
@@ -173,8 +172,12 @@ def iiif_service(id)
     )
   end
 
+  def all_dc_nodes
+    public_xml_document.xpath '//oai_dc:dc/*', 'dc' => DC_SCHEMA, 'oai_dc' => OAI_DC_SCHEMA
```
cbeer: Is the `dc` namespace doing something for us here?

app/models/iiif_presentation_manifest.rb
```diff
@@ -173,8 +172,12 @@ def iiif_service(id)
...
+  end
+
   def get_metadata(label, xpath)
```
cbeer: Is the publish date the only consumer of this method now? Maybe that suggests a refactoring, at least to remove the unused namespace elements, but maybe this logic would be just as clear inline?

app/models/iiif_presentation_manifest.rb
```diff
@@ -69,13 +72,9 @@ def body(controller = nil)
...
 
-    metadata = []
-    # make into method, pass in xpath and label
-    metadata += get_metadata 'Creator', '//oai_dc:dc/dc:creator'
-    metadata += get_metadata 'Contributor', '//oai_dc:dc/dc:contributor'
-    metadata += get_metadata 'Publisher', '//oai_dc:dc/dc:publisher'
-    metadata += get_metadata 'Date', '//oai_dc:dc/dc:date'
-    metadata += get_metadata 'PublishVersion', '/publicObject/@publishVersion'
+    # add all DC metadata to metadata section as key value pairs
+    metadata = all_dc_nodes.map { |dc_node| { 'label' => dc_node.name.upcase_first, 'value' => dc_node.text } }
```
cbeer: I wonder if this logic would be easier to isolate and test if it was pushed into its own method? Maybe as `dc_to_iiif_metadata` or something like that?

.rubocop_todo.yml
peetucket: actually, i think this is unnecessary anyway -- it got added automatically when i did a new autogen, but i believe the exclusion is already in the .rubocop.yml

app/models/iiif_presentation_manifest.rb
peetucket: i presume you mean the get_metadata method?  although both of those methods are now only used once, and could be wrapped up and used in a single dc_to_iiif_metadata method as you note below

app/models/iiif_presentation_manifest.rb
peetucket: i don't know xquery well enough to understand why it was there in the first place (it was there before) but it works with just OAI_DC namespace, so I will remove the DC reference

app/models/iiif_presentation_manifest.rb
peetucket: refactored to a new method

app/models/iiif_presentation_manifest.rb
```diff
@@ -69,15 +71,7 @@ def body(controller = nil)
...
-    metadata = []
-    # make into method, pass in xpath and label
-    metadata += get_metadata 'Creator', '//oai_dc:dc/dc:creator'
-    metadata += get_metadata 'Contributor', '//oai_dc:dc/dc:contributor'
-    metadata += get_metadata 'Publisher', '//oai_dc:dc/dc:publisher'
-    metadata += get_metadata 'Date', '//oai_dc:dc/dc:date'
-    metadata += get_metadata 'PublishVersion', '/publicObject/@publishVersion'
-
-    manifest.metadata = metadata if metadata.present?
+    manifest.metadata = dc_to_iiif_metadata if dc_to_iiif_metadata.present?
```
jmartin-sul: i'm not sure how expensive it is to compute this, but it seems like it might be a good idea to cache the return value of `dc_to_iiif_metadata` in a var, so as to not run through its operations twice?  def wouldn't hold up the PR over it, but seemed worth suggesting.

app/models/iiif_presentation_manifest.rb
```diff
@@ -6,6 +6,8 @@ class IiifPresentationManifest
 
   attr_reader :purl_resource
 
+  OAI_DC_SCHEMA = 'http://www.openarchives.org/OAI/2.0/oai_dc/'.freeze
```
peetucket: this needs to match what is in the DC datastream itself or else the xpath query in the get_metadata method won't work... which makes it fragile actually.  is there a better way to search the xml to avoid this?

jmartin-sul: ah, ok, that makes sense.  if it's just being used for matching, i think it's fine.  i don't know one way or the other whether a wildcard would work there.  i was just reflexively trying to encourage less usage of unsecure network connections, but since there's no connection being made here, not an issue.

would be nice if the datastreams used https from here on out, but that's obviously well outside the scope of this PR.  if you had both, i guess you could have each version of the URL, and just use the results of running both xpath queries.  there's probably a less kludgy way of doing that, but seems like we can cross that bridge if/when we come to it.

app/models/iiif_presentation_manifest.rb
```diff
@@ -173,16 +167,20 @@ def iiif_service(id)
...
-      {
-        'label' => label,
-        'value' => node.text
-      }
+  # transform all DC metadata in the public XML into an array of hashes for inclusion in the IIIF manifest
+  def dc_to_iiif_metadata
+    @dc_to_iiif_metadata ||= begin
+      all_dc_nodes = public_xml_document.xpath '//oai_dc:dc/*', 'oai_dc' => OAI_DC_SCHEMA
+      metadata = all_dc_nodes.map { |dc_node| iiif_key_value(dc_node.name.upcase_first, dc_node.text) }
+      metadata += public_xml_document.xpath('/publicObject/@published').map { |node| iiif_key_value('aPublishDate', node.text) } # add published date
```
jmartin-sul: looks like it's just this stray `a` in `'aPublishDate'` leading to the new test failure

app/models/iiif_presentation_manifest.rb
peetucket: memorized in the method

PR #123, opened 2017-04-11
Use the original_url for the IIIF manifest's @id value.

### LINKED ISSUE #122: IIIF Manifests @id can be non spec compliant.
According to the [IIIF spec](http://iiif.io/api/presentation/2.1/#id)

```
A manifest must have exactly one id, and it must be the http(s) URI at which it is published.
```

We allow our the IIIF manifest to be returned under the following URLs:
- https://purl.stanford.edu/fw090jw3474/iiif/manifest
- https://purl.stanford.edu/fw090jw3474/iiif/manifest.json

However; the URL returned in the `@id` parameter is always `https://purl.stanford.edu/fw090jw3474/iiif/manifest`.  This is causing some issues with some 3rd party software that implements the IIIF spec.

One possibility would be to use `controller.request.original_url` instead of `iiif_manifest_url` [here](https://github.com/sul-dlss/purl/blob/master/app/models/iiif_presentation_manifest.rb#L56)

If that sounds agreeable, I'm happy to submit a PR with that change (as this could potentially cause some issues with a demo we have next week if somebody adds the `.json` in their manifest URLs).

jkeck:
Fixes #122

### DIFF
--- a/app/controllers/purl_controller.rb
+++ b/app/controllers/purl_controller.rb
@@ -50,7 +50,7 @@ class PurlController < ApplicationController
     return unless stale?(last_modified: @purl.updated_at.utc, etag: @purl.cache_key + "/#{@purl.updated_at.utc}")
 
     if @purl.iiif_manifest?
-      manifest = Rails.cache.fetch([@purl, @purl.updated_at.utc], expires_in: Settings.resource_cache.lifetime) do
+      manifest = Rails.cache.fetch([@purl, params[:format], @purl.updated_at.utc], expires_in: Settings.resource_cache.lifetime) do
         @purl.iiif_manifest.body(self).to_ordered_hash
       end
 
--- a/spec/model/iiif_presentation_manifest_spec.rb
+++ b/spec/model/iiif_presentation_manifest_spec.rb
@@ -1,8 +1,69 @@
 require 'rails_helper'
 
+class TestPurlController
+  delegate :iiif_manifest_url, :purl_url, to: :helpers
+  delegate :original_url, to: :request
+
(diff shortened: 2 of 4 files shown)


spec/model/iiif_presentation_manifest_spec.rb
```diff
@@ -1,9 +1,70 @@
 require 'rails_helper'
 
+class TestPurlController
```
jkeck: Couldn't figure out a better way to mock out a controller context here.  I've also added a feature test FWIW.

app/controllers/purl_controller.rb
```diff
@@ -50,7 +50,7 @@ def manifest
     return unless stale?(last_modified: @purl.updated_at.utc, etag: @purl.cache_key + "/#{@purl.updated_at.utc}")
 
     if @purl.iiif_manifest?
-      manifest = Rails.cache.fetch([@purl, @purl.updated_at.utc], expires_in: Settings.resource_cache.lifetime) do
+      manifest = Rails.cache.fetch([@purl, params[:format], @purl.updated_at.utc], expires_in: Settings.resource_cache.lifetime) do
```
jkeck: This is what might lead me to think we should just enforce a particular route.  Given that we cache the manifest we might end up keep two cached versions of the same manifest with the only difference being the `@id`.

¯\\_(ツ)_/¯

jkeck:
Closing this as it's clear from the thread on #122 indicates that this is not just a problem w/ `.json` and is a larger issue w/ how extensions are handled for the manifest URLs.

PR #125, opened 2017-04-27
Require bare manifest urls and redirect manifest.json requests to the…

cbeer:
… bare urls

Fixes https://github.com/sul-dlss/purl/pull/122


ndushay:
👍 to merge.   Does similar code need to be added to v3 manifests?  (as a sep issue)?

cbeer:
Sure, I'll do that now.

cbeer:
Not pedantic at all; it should have tests.

PR #143, opened 2017-06-23
plumbing for v3uv manifest creation

### LINKED ISSUE #138: add plumbing for IIIF v3.0 controller / routes etc.
WIP branch:  iiifv3-plumbing

part of #133

ndushay:
Resolves #138

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,7 +22,7 @@ Metrics/AbcSize:
   Max: 22
 
 Metrics/ClassLength:
-  Max: 200
+  Max: 205
 
 Metrics/MethodLength:
   Max: 20
--- a/app/controllers/iiif_v3_controller.rb
+++ b/app/controllers/iiif_v3_controller.rb
@@ -1,0 +1,55 @@
+class IiifV3Controller < ApplicationController
+  before_action :load_purl
+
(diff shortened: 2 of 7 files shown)


.rubocop.yml
```diff
@@ -22,7 +22,7 @@ Metrics/AbcSize:
   Max: 22
 
 Metrics/ClassLength:
-  Max: 200
+  Max: 205
```
ndushay: new methods added to `app/models/purl_resource.rb` made it 5 too long.

app/controllers/iiif_v3_controller.rb
```diff
@@ -0,0 +1,55 @@
+class IiifV3Controller < ApplicationController
```
ndushay: lots of duplicate code here, but I'm thinking it will be better to have the two controllers be separate (?)

app/models/iiif3_presentation_manifest.rb
```diff
@@ -0,0 +1,98 @@
+require 'iiif/presentation'
+
+class Iiif3PresentationManifest < IiifPresentationManifest
```
ndushay: here there is so much duplicated code, I went with inheritance.  As we fully implement v3 manifests, we may want to revisit this.

Issue #133, opened 2017-06-01
Expose v3 presentation manifests as application routes


cbeer:
- [x] rails plumbing (see #138)
- [x] actually serve v3 presentation manifests for
  - [x] pdf
  - [x] 3d object

ndushay:
@cbeer   is this accomplished now?  Would we make separate tickets remaining stuff on here if we want to do it?

PR #149, opened 2017-06-28
Update to produce IIIF v3-like manifests

### LINKED ISSUE #147: implement iiif3 manifests for pdf objects
related to #132

We have iiif v3 presentation fixture objects for pdfs  from Ben.  (in `spec/fixtures`).

- [ ] We need the purl code to create v3 manifests that include pdfs.

This relies on sul-dlss/osullivan#1 and sul-dlss/osullivan#4.

- [ ] allow `id` and `@id` to work OR allow us to inject the iiif v2 service in osullivan

cbeer:
closes #147

Fixes https://github.com/sul-dlss/purl/issues/133.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,7 +22,10 @@ Metrics/AbcSize:
   Max: 22
 
 Metrics/ClassLength:
-  Max: 205
+  Exclude:
+    - app/models/purl_resource.rb
+    - app/models/iiif_presentation_manifest.rb
+    - app/models/iiif3_presentation_manifest.rb
 
 Metrics/MethodLength:
   Max: 20
--- a/Gemfile
+++ b/Gemfile
@@ -45,7 +45,7 @@ gem 'addressable'
 gem 'tophat'
 gem 'rails-file-icons'
 gem 'sul_styles', '~> 0.3'
-gem 'iiif-presentation'
+gem 'iiif-presentation', github: 'sul-dlss/osullivan', branch: 'uv-v3'
 gem 'dalli'
 
 group :production do
--- a/Gemfile
+++ b/Gemfile
@@ -45,7 +45,9 @@ gem 'addressable'
 gem 'tophat'
 gem 'rails-file-icons'
 gem 'sul_styles', '~> 0.3'
-gem 'iiif-presentation', github: 'sul-dlss/osullivan', branch: 'uv-v3'
+
+# sul-dlss/osullivan#development has early support for generating IIIF v3 manifests
+gem 'iiif-presentation', github: 'sul-dlss/osullivan', branch: 'development'
 gem 'dalli'
 
 group :production do
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,10 @@
 source 'https://rubygems.org'
 
+git_source(:github) do |repo_name|
+  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
+  "https://github.com/#{repo_name}.git"
+end
+
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.0.1'
 # Use sqlite3 as the database (during local development)
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -2,14 +2,15 @@ require 'iiif/presentation'
 require 'iiif/v3/presentation'
 
 class Iiif3PresentationManifest < IiifPresentationManifest
-  # Bypass this method if there are no image resources in contentMetadata
+  delegate :resources, to: :content_metadata
+
   # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
   def body(controller = nil)
     controller ||= Rails.application.routes.url_helpers
     purl_base_uri = controller.purl_url(druid)
 
     manifest_data = {
-      'id' => controller.iiif3_manifest_url(druid), # DIFF from v2
+      'id' => controller.iiif3_manifest_url(druid),
       'label' => title,
       'attribution' => copyright || 'Provided by the Stanford University Libraries',
       'logo' => {
@@ -19,11 +20,7 @@ class Iiif3PresentationManifest < IiifPresentationManifest
       'seeAlso' => {
         'id' => controller.purl_url(druid, format: 'mods'),
         'format' => 'application/mods+xml'
-      },
-      '@context' => [
-        'http://www.w3.org/ns/anno.jsonld',
-        'http://iiif.io/api/presentation/3/context.json'
-      ]
+      }
     }
 
     manifest = IIIF::V3::Presentation::Manifest.new manifest_data
@@ -43,10 +40,7 @@ class Iiif3PresentationManifest < IiifPresentationManifest
     manifest.thumbnail = thumbnail_resource
 
     # for each resource image, create a canvas
-    page_images.each do |resource|
-      next unless purl_resource.rights.world_rights_for_file(resource.filename).first ||
-                  purl_resource.rights.stanford_only_rights_for_file(resource.filename).first
-
+    resources.each do |resource|
       sequence.canvases << canvas_for_resource(purl_base_uri, resource)
     end
 
@@ -56,11 +50,13 @@ class Iiif3PresentationManifest < IiifPresentationManifest
 
   def canvas_for_resource(purl_base_uri, resource)
     canv = IIIF::V3::Presentation::Canvas.new
-    canv['id'] = "#{purl_base_uri}/iiif3/canvas/#{resource.id}" # DIFF from v2
+    canv['id'] = "#{purl_base_uri}/iiif3/canvas/#{resource.id}"
     canv.label = resource.label
     canv.label = 'image' unless canv.label.present?
-    canv.height = resource.height
-    canv.width = resource.width
+    if image?(resource)
+      canv.height = resource.height
+      canv.width = resource.width
+    end
 
     anno_page = IIIF::V3::Presentation::AnnotationPage.new
     anno_page['id'] = "#{purl_base_uri}/iiif3/annotation_page/#{resource.id}"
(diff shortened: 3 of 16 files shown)


ndushay:
The three concerns mentioned above (as checkboxes) are reflected in sul-dlss/osullivan#8

Gemfile
```diff
@@ -40,7 +40,7 @@ gem 'addressable'
 gem 'tophat'
 gem 'rails-file-icons'
 gem 'sul_styles', '~> 0.3'
-gem 'iiif-presentation'
+gem 'iiif-presentation', github: 'sul-dlss/osullivan', branch: 'uv-v3'
```
ndushay: Please add a comment explaining why we're not using a released gem.

actually, should we wait for the merge to develop and use the develop branch?  (The current correct branch is `uv-v3-new` in any case, due to rebasing woes)

app/models/iiif3_presentation_manifest.rb
```diff
@@ -8,20 +9,24 @@ def body(controller = nil)
     purl_base_uri = controller.purl_url(druid)
 
     manifest_data = {
-      '@id'   => controller.iiif3_manifest_url(druid), # DIFF from v2
+      'id' => controller.iiif3_manifest_url(druid), # DIFF from v2
```
ndushay: I wonder if we want to get rid of all the `# DIFF from v2` comments, or add them to all the important lines.  The way it is now is misleading, IMO.

app/models/iiif3_presentation_manifest.rb
```diff
@@ -50,27 +55,33 @@ def body(controller = nil)
...
 
-    anno = IIIF::Presentation::Annotation.new
-    anno['@id'] = "#{purl_base_uri}/iiif3/annotation/#{resource.id}" # DIFF from v2
+    anno = IIIF::V3::Presentation::Annotation.new
+    anno['id'] = "#{purl_base_uri}/iiif3/annotation/#{resource.id}" # DIFF from v2
 
-    img_res = IIIF::Presentation::ImageResource.new
-    img_res['@id'] = "#{url}/full/full/0/default.jpg"
+    img_res = IIIF::V3::Presentation::ImageResource.new
+    img_res['type'] = 'Image'
```
ndushay: does this line assigning `type` go away once we implement sul-dlss/osullivan#5 ?  if so, I would indicate that in a comment.

app/models/iiif3_presentation_manifest.rb
```diff
@@ -1,4 +1,5 @@
 require 'iiif/presentation'
+require 'iiif/v3/presentation'
```
ndushay: I'm starting to believe we might indicate with a comment that we are using inheritance but if this class grows to override a large percentage of the v2 class, then we should reconsider.

Gemfile
```diff
@@ -40,7 +45,7 @@ gem 'addressable'
 gem 'tophat'
 gem 'rails-file-icons'
 gem 'sul_styles', '~> 0.3'
-gem 'iiif-presentation'
+gem 'iiif-presentation', github: 'sul-dlss/osullivan', branch: 'development'
```
ndushay: I would still like a comment indicating why we aren't using a released gem ...

app/models/iiif3_presentation_manifest.rb
ndushay: I take it back:  here.  Indicate with a comment that we are using inheritance but if this class grows to override a large percentage of the v2 class, then we should reconsider.

app/models/purl_resource.rb
```diff
@@ -101,10 +101,6 @@ def iiif3_manifest
     @iiif3_manifest ||= Iiif3PresentationManifest.new(self)
   end
 
-  def iiif3_manifest?
-    iiif3_manifest.needed?
-  end
-
```
ndushay: does this mean we can reduce max class size in rubocop from 205 to 201?

spec/features/iiif3_manifest_spec.rb
```diff
@@ -84,14 +84,6 @@
...
   end
 
-  it 'suppresses sequences for dark resources' do
-    visit '/bc421tk1152/iiif3/manifest'
-    json = JSON.parse(page.body)
-
-    expect(json['sequences'].length).to eq 1
-    expect(json['sequences'].first['canvases']).to be_blank
-  end
-
```
ndushay: did this functionality change?  Do we no longer suppress sequences for dark resources?  (I barely know what that means)

do we have a test indicating the new behavior?

spec/integration/purl_spec.rb
```diff
@@ -49,10 +49,6 @@
         expect(json_body['label']).to eq('John Wyclif and his followers, Tracts in Middle English')
         expect(json_body['metadata'].length).to eq 16
       end
-      it 'renders nil for a non-manifest' do
-        visit "/#{@file_object}/iiif3/manifest"
-        expect(page.status_code).to eq(404)
-      end
```
ndushay: has this changed?  why?  do we have a test indicating the new behavior?

spec/integration/purl_spec.rb
```diff
@@ -189,15 +185,5 @@
...
-    it '/status/all has response code 200' do
-      visit '/status/all'
-      expect(page.status_code).to eq 200
-      expect(page).to have_text('HTTP check successful')
-    end
-    it '/status/feedback_mailer responds' do
-      visit '/status/feedback_mailer'
-      expect(page.status_code).to eq 200
-      expect(page).to have_text('FeedbackMailer')
-    end
```
ndushay: why are we removing these checks?

spec/features/iiif3_manifest_spec.rb
```diff
@@ -165,28 +159,55 @@
...
+      expect(canvas['content'].first['items'].length).to eq 1
+      expect(canvas['height']).not_to be_present
+      expect(canvas['width']).not_to be_present
+
+      pdf = canvas['content'].first['items'].first
+      expect(pdf['body']['id']).to eq 'https://stacks.stanford.edu/file/bb132pr2055/Puente-Thesis-Submission-augmented.pdf'
+      expect(pdf['body']['format']).to eq 'application/pdf'
+      expect(pdf['body']['type']).to eq 'Document'
+    end
+  end
```
ndushay: if we're saying "3d objects come for free", should we have a 3D object spec here, just to prove it?  I'm not sure inebriated coder will know that pdf object = 3d object in iiif manifest land.

spec/integration/purl_spec.rb
cbeer: We ran into these  after disabling caching for the test environment caused the `/all` test to fail. They seem like it's testing behavior from upstream in okcomputer, no? I left the one that proves we've wired it in correctly.

spec/integration/purl_spec.rb
cbeer: That has changed. With IIIF v3, we're generating IIIF manifests for all types of objects. The PDF test we added demonstrates that.

spec/features/iiif3_manifest_spec.rb
cbeer: Once we have an appropriate 3D object, we could do that. Right now, none of our "3d" objects have a 3js-flavored file.

spec/features/iiif3_manifest_spec.rb
cbeer: Yes (and no one really seems to really know what that means.. it's not like it was suppressing the files from the public xml).

I'm not sure it's necessary to test that we don't do something, is it?

app/models/purl_resource.rb
cbeer: Sure; instead of continuing to bump that size up and down, I'm just going to exclude this file (and our IIIF generators) so we can just adopt the rubocop's default size instead.

app/models/iiif3_presentation_manifest.rb
cbeer: We're already overriding 75% of the lines of code from `IiifPresentationManifest`. Maybe that time is now (or after merging this and getting a 👍 that we're even doing IIIF v3 right)?

ndushay: I'm fine with it becoming a separate ticket to do.  I just didn't want my rationale and its evaluation to be left standing if and when it doesn't make sense.

spec/features/iiif3_manifest_spec.rb
ndushay: only need to test it if we're trying to document current behavior.  Sounds like this is NOT one of those cases, so fine with me.

spec/features/iiif3_manifest_spec.rb
ndushay: 1.  don't we have one 3js file for one object?  Is that not accessioned as part of the object yet?
2.  perhaps make a ticket about adding such a test if we have no ability to write such a test now?  (I would argue that perhaps the test could be written now, as it is checking the manifest, not the data?)

spec/integration/purl_spec.rb
ndushay: I'm only looking at the snippet ... are we leaving this test for v2 iiif and only removing it for v3 ?

.rubocop.yml
```diff
@@ -22,7 +22,10 @@ Metrics/AbcSize:
   Max: 22
 
 Metrics/ClassLength:
-  Max: 205
+  Exclude:
+    - app/models/purl_resource.rb
+    - app/models/iiif_presentation_manifest.rb
+    - app/models/iiif3_presentation_manifest.rb
```
ndushay: 👍 for switching to exclude for class length in rubocop

Gemfile
```diff
@@ -40,7 +45,9 @@ gem 'addressable'
 gem 'tophat'
 gem 'rails-file-icons'
 gem 'sul_styles', '~> 0.3'
-gem 'iiif-presentation'
+
+# sul-dlss/ossullivan#development has early support for generating IIIF v3 manifests
```
ndushay: thank you for comment in Gemfile.  It's not worth twiddling if this PR is otherwise ready, but `osssssssullivan`  (there's one extra s)

app/models/iiif3_presentation_manifest.rb
cbeer: 👌 https://github.com/sul-dlss/purl/issues/151.

ndushay: I think I did +1 on making an issue for refactoring `iiif3_presentation_manifest` on my earlier review.  (and it was my bad for not putting in such a comment when I set up this class about 5 PRs ago)

spec/integration/purl_spec.rb
cbeer: Yes. We're not changing v2 manifest creation at all.

cbeer:
I've made the issue for refactoring and left it on your last comment about that. https://github.com/sul-dlss/purl/issues/151

Gemfile
cbeer: When this is ready to go, I'll rebase and squash e64f6f7 into the right place.

spec/features/iiif3_manifest_spec.rb
jmartin-sul: i'd vote for option 2, if that's ok.  the distinction (in my limited understanding) is image vs not image, so it seems like PDF should be a fine example for not image.   but i like comprehensive tests, so maybe once we have a 3js example, we can more easily add a 3js test, without going out of our way to create an example object now.  i'm assuming 3js is one of the object types @LynnMcRae and @blalbrit are working up in the set of examples they're creating?

cbeer: Ticket filed: https://github.com/sul-dlss/purl/issues/152

PR #153, opened 2017-06-29
Pretty-print our JSON output

cbeer:


cbeer:
I'm tired of having to pretty-print our responses manually, and this'll make it easier to share with the IIIF community.

PR #156, opened 2017-06-30
Expose IIIF v3 annotation pages as routes; fixes #154

### LINKED ISSUE #154: Expose AnnotationPage response as its own route
From the spec:

> An AnnotationPage must have exactly one id, and it must be the http(s) URI at which it is published.

http://prezi3.iiif.io/api/presentation/3.0/#technical-properties

cbeer:


cbeer:
I think it's already covered. That line just moved from `canvas_for_resource` to `annotation_for_resource`.

PR #163, opened 2017-07-11
Authentication service description for location restricted materials

### LINKED ISSUE #134: Add authentication service description for location-based restrictions
(Presumably based on http://iiif.io/api/auth/1.0/#external-interaction-pattern?)

We've made this change in stacks, but it would be a courtesy to consuming clients to also expose this information in our presentation manifests.

tingulfsen:
Fixes #134 

This also makes a minor change to regular access restrictions for images.

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -109,9 +109,13 @@ class Iiif3PresentationManifest < IiifPresentationManifest
     img_res.width = resource.width
 
     img_res.service = iiif_image_v2_service(url)
+    img_res.service['service'] = []
+    if purl_resource.rights.stanford_only_rights_for_file(resource.filename).first
+      img_res.service['service'].append(iiif_stacks_login_service)
+    end
 
-    unless purl_resource.rights.world_rights_for_file(resource.filename).first
-      img_res.service['service'] = [iiif_stacks_login_service]
+    if purl_resource.rights.restricted_by_location?(resource.filename)
+      img_res.service['service'].append(iiif_location_auth_service)
     end
 
     img_res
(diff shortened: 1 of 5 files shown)


app/models/iiif3_presentation_manifest.rb
```diff
@@ -101,11 +101,15 @@ def image_resource(resource)
     img_res.width = resource.width
 
     img_res.service = iiif_image_v2_service(url)
-
-    unless purl_resource.rights.world_rights_for_file(resource.filename).first
+    if purl_resource.rights.stanford_only_rights_for_file(resource.filename)
```
cbeer: I wonder if this should be `.first`? like the `world_rights_for_file` was? If so, it's kinda worrying the tests didn't catch it.. If not, it's annoying that the methods aren't parallels..

tingulfsen: Good catch - I'll investigate.

app/models/iiif3_presentation_manifest.rb
```diff
@@ -144,13 +148,30 @@ def iiif_image_v2_service(id)
...
+      ]
+    )
+  end
+
+  def iiif_location_auth_service
+    IIIF::V3::Presentation::Service.new(
+      'profile' => 'http://iiif.io/api/auth/1/external',
+      'label' => 'External Authentication Required',
+      'failureHeader' => 'Restricted Material',
+      'failureDescription' => 'Restricted content cannot be accessed from your location'
```
cbeer: We should get https://github.com/sul-dlss/stacks/pull/138 merged, and then update this to include the token api.

app/models/iiif3_presentation_manifest.rb
```diff
@@ -144,13 +148,30 @@ def iiif_image_v2_service(id)
...
-        'id' => "#{Settings.stacks.url}/image/iiif/token",
-        'profile' => 'http://iiif.io/api/auth/1/token'
-      }]
+      'service' => [
+        {
+          'id' => "#{Settings.stacks.url}/image/iiif/token",
+          'profile' => 'http://iiif.io/api/auth/1/token'
+        },
+        {
+          '@id' => "#{Settings.stacks.url}/auth/logout",
```
cbeer: I forget which is these should be `id` vs `@id`.

tingulfsen: It looks like all the examples at http://iiif.io/api/auth/1.0/ are `@id`.

cbeer: Ok. We should fix the bare `id` above then, huh?

app/models/iiif3_presentation_manifest.rb
tingulfsen: 👍

app/models/iiif3_presentation_manifest.rb
tingulfsen: Sorry @cbeer you are too polite...I didn't realize that was what you were getting at. I have fixed the offender.

app/models/iiif3_presentation_manifest.rb
```diff
@@ -101,11 +101,15 @@ def image_resource(resource)
     img_res.width = resource.width
 
     img_res.service = iiif_image_v2_service(url)
-
-    unless purl_resource.rights.world_rights_for_file(resource.filename).first
+    img_res.service['service'] = []
+    if purl_resource.rights.stanford_only_rights_for_file(resource.filename).first
       img_res.service['service'] = [iiif_stacks_login_service]
```
jmartin-sul: since `img_res.service['service']` is initialized to a list, maybe this could be changed to `img_res.service['service'].append([iiif_stacks_login_service])`, just so it mirrors the construction below?  no practical behavioral difference, and that's pretty nitpicky, so i'd also be fine to leave it as-is.

app/models/iiif3_presentation_manifest.rb
jmartin-sul: does this still need doing before this PR is merged?

cbeer: Yes, I think so.

Issue #145, opened 2017-06-26
"publish" v3 iiif manifests for "pdf objects"


ndushay:
- [ ] how do we know when we have a "pdf object"?   Is it a non-image, non-media object with a pdf?   What if there are images and pdf, or media and pdf?  or xxx and pdf?

- [ ] purl app uses iiif3  routes for ^^ object presentation manfests

is this covered in #133 ?

cbeer:
> how do we know when we have a "pdf object"? 

Maybe a better way to rephrase the task is "include PDFs in the v3 IIIF manifests"? We shouldn't need to care if it is a "PDF object" or not.

PURL doesn't consume the IIIF manifests either -- that's done over in sul-embed.

ndushay:
okay, maybe I don't understand this well enough, but "implement v3  (UV flavor)  iiif  manifests for everything" seems like a bigger task than focusing on one type of data first.

blalbrit:
per @LynnMcRae (https://consul.stanford.edu/pages/viewpage.action?title=DOR+content+types%2C+resource+types+and+interpretive+metadata&spaceKey=chimera)

Proposed content types (in progress)
for content-type=document, resource types are:
document – must contain a file that is supported for viewing, starting with PDF (any requirements, limitations?).  Other types may be added in the future. 
alternative model: distinguish resource-type to correspond with different types of docs.  
for content-type=3d, resource types are:
3d -- in progress, samples files at http://web.stanford.edu/~cengel/3D/bf418ns2590/

blalbrit:
note ^^ have not been implemented yet, but we can do with the fixture objects if @LynnMcRae says this is stable enough to proceed with.

jmartin-sul:
duplicates #147

feel free to reopen if that seems incorrect

PR #176, opened 2017-07-27
iiif sequence (v2 and v3): populate viewingDirection correctly

ndushay:
I stumbled on this while writing tests for osullivan that mimic our purl code as exactly as is practical.

Basically, the osullivan code uses underscores for the true keys of its hashes.  It is full of cleverness so that you can use different forms of accessors to get the data.  For example, both of these statements should pass:
```ruby
          expect(sequence_vd.viewing_direction).to eq 'left-to-right'
          expect(sequence_vd.viewingDirection).to eq 'left-to-right'
```
For this to be true, however, the underlying hash is expecting underscored keys.  So: 
```ruby
sequence_object['viewing_direction'] = 'left-to-right' # works for both
sequence_object['viewingDirection'] = 'left-to-right' # only works for .viewingDirection
sequence_object.viewingDirection = 'left-to-right' # works for both
sequence_object.viewing_direction = 'left-to-right' # works for both
```

Thus, @tingulfsen managed to pick the ONE WAY out of four that isn't appropriate for osullivan:  do NOT have camelcase keys as a direct hash method.

I also reopened #145  as I had an impression that Lynn wanted "left-to-right" to be explicitly added to the objects when there was no other clue about them.  That is not part of this PR ... but it is the same part of the code.

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -43,7 +43,7 @@ class Iiif3PresentationManifest < IiifPresentationManifest
       'label' => 'Current order'
     )
 
-    sequence['viewingDirection'] = VIEWING_DIRECTION[order] if order
+    sequence.viewingDirection = VIEWING_DIRECTION[order] if order
 
     manifest.thumbnail = [thumbnail_resource]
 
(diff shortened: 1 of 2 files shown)


ndushay:
@cbeer @jmartin-sul   the explanation above might interest you, but I think Tommy should be the reviewer ...

PR #178, opened 2017-08-04
Default viewingDirection should be left-to-right

### LINKED ISSUE #165: Presentation manifest for book/manuscript should include viewingDirection
Content Metadata for type=book can include a <bookData> tag with an attribute to indicate reading direction, to wit (https://purl.stanford.edu/yr183sf1341):

```
<contentMetadata type="book" objectId="yr183sf1341">
     <bookData readingOrder="rtl"/>
```
"ltr" is the default, likely never to be explicitly expressed. "rtl" as shown here should translate to a `<viewingDirection>` value of "right-to-left" in the presentation manifest for books (or manuscripts).

tingulfsen:
Fixes #165 (once more).

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -4,11 +4,6 @@ require 'iiif/v3/presentation'
 class Iiif3PresentationManifest < IiifPresentationManifest
   delegate :reading_order, :resources, to: :content_metadata
 
-  VIEWING_DIRECTION = {
-    'ltr' => 'left-to-right',
-    'rtl' => 'right-to-left'
-  }.freeze
-
   # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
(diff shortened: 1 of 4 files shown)


app/models/iiif3_presentation_manifest.rb
```diff
@@ -43,7 +38,12 @@ def body(controller = nil)
       'label' => 'Current order'
     )
 
-    sequence.viewingDirection = VIEWING_DIRECTION[order] if order
+    sequence.viewingDirection = case order
+                                when nil
+                                  VIEWING_DIRECTION['ltr']
```
ndushay: where is VIEWING_DIRECTION  defined?  It seems to be removed ...

tingulfsen: @jmartin-sul suggested using the definition in the superclass so we don't have to duplicate the code - I went with that.

PR #182, opened 2017-09-21
IIIF sequences should include the representative thumbnail image, eve…

### LINKED ISSUE #181: Should we modify the logic for how images are returned in a sequence?
Currently our logic only returns files if they are "deliverable": https://github.com/sul-dlss/purl/blob/master/app/models/iiif_presentation_manifest.rb#L34

That logic being:

Stanford only rights OR World rights.

This logic _seems_ like it is ok, but I would offer it may be problematic in a case where we _can_ offer an image as metadata for 400 x 400. 

Examples of these types of objects are plentiful, but I will focus on https://purl.stanford.edu/zx774pr7698 as it is a map image from DRMC, which should be at least thumb viewable, no?

We are showing these in SearchWorks as thumbs (via Stacks delivery).. why can't we offer them as limited support as resources in the manifest?

cc @snydman

cbeer:
…n if it is not otherwise available

Closes #181

### DIFF
--- a/app/models/iiif_presentation_manifest.rb
+++ b/app/models/iiif_presentation_manifest.rb
@@ -42,7 +42,8 @@ class IiifPresentationManifest
   # also is this right?  should it handle location rights too?
   def deliverable_file?(file)
     purl_resource.rights.stanford_only_rights_for_file(file.filename).first ||
-      purl_resource.rights.world_rights_for_file(file.filename).first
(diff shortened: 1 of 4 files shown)


mejackreed:
Waiting on approval from @snydman && @LynnMcRae

mejackreed:
Approved.

PR #197, opened 2017-12-14
Adds view in EarthWorks link for geo things

### LINKED ISSUE #189: Add "View in EarthWorks" for Geo objects

mejackreed:
Closes #189 

## cg357zz0321

screen shot 2017-12-14 at 8 20 54 am

### DIFF
--- a/app/views/purl/_find_it.html.erb
+++ b/app/views/purl/_find_it.html.erb
@@ -2,9 +2,16 @@
   <%= link_to 'View in SearchWorks', Settings.searchworks.url + @purl.catalog_key %>
 <% end %>
 
+<% releases = [] %>
+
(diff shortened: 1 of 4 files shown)


jkeck:
This looks good to me.  Worth adding a test?

mejackreed:
Possibly. I think the reason I didn't was that PURL specs have pear tree representation of objects in fixtures. I'm happy to add content type geo if that makes sense here. I held off because it would be  a bigish change.

jkeck:
Ah, yeah, totally understand, that seems like a PITA for such a small change.  Would a view spec be easy enough to write or still a pain?  If so, I'm fine with 🚢ing it.

mejackreed:
Thanks!

PR #203, opened 2018-01-22
Advertise the IIIF location auth service for appropriate resources

cbeer:


app/models/iiif_presentation_manifest.rb
```diff
@@ -188,24 +188,14 @@ def annotation_for_resource(purl_base_uri, resource)
...
-            'profile' => 'http://iiif.io/api/auth/1/token'
-          }]
-        )
-      ]
+    if purl_resource.rights.stanford_only_rights_for_file(resource.filename).first
+      img_res.service['service'] = [iiif_stacks_login_service]
+    end
+
+    if purl_resource.rights.restricted_by_location?(resource.filename)
+      img_res.service['service'].append(iiif_location_auth_service)
```
jkeck: Just a quick thought here, but it looks like w went from "any scenario where the file was not world accessible" to "only these two scenarios where the file is restricted".

Our rights are kind of confusing, so I'm guessing that this covers all scenarios we might want it? I wonder if there are any others that we need to be concerned about.  If we add new scenarios, they'll need to be accounted for here, but perhaps that would have been true anyway.

cbeer: Yes, we think it covers the various scenarios. The code before, although seeming to account for other possibilities, was actually limited to "things that webauth would fix" due to the auth implementation behind it, so it's not a big change (also, this was plucked directly from the IIIF v3 implementation, which was somewhat vetted, at least)

PR #205, opened 2018-01-22
DRY IIIF v2 + v3 controllers

cbeer:


app/controllers/iiif_v2_controller.rb
```diff
@@ -4,14 +4,21 @@ class IiifV2Controller < ApplicationController
...
   def manifest
     return unless stale?(last_modified: @purl.updated_at.utc, etag: @purl.cache_key + "/#{@purl.updated_at.utc}")
 
-    if @purl.iiif_manifest?
-      manifest = Rails.cache.fetch([@purl, @purl.updated_at.utc, 'iiif_v2', 'manifest'], expires_in: Settings.resource_cache.lifetime) do
-        @purl.iiif_manifest.body(self).to_ordered_hash
-      end
+    cache_key = [@purl, @purl.updated_at.utc, 'iiif', iiif_version, 'manifest']
+    manifest = Rails.cache.fetch(cache_key, expires_in: Settings.resource_cache.lifetime) do
+      @purl.iiif_manifest(iiif_version).body(self).try(:to_ordered_hash)
```
jcoyne: Why is the `try` here?  Is there some time `body()` would return `nil`?

app/controllers/iiif_v3_controller.rb
```diff
@@ -1,52 +1,10 @@
...
-    end
-
-    if manifest
-      render json: JSON.pretty_generate(manifest.as_json)
-    else
-      head :not_found
-    end
-  end
-
+class IiifV3Controller < IiifV2Controller
```
jcoyne: I'd rather see a `IiiifAbstractController`. This inheritance seems awkward.

jcoyne:
We no longer have separate v2 and v3 controllers, so closing this.

Issue #269, opened 2018-11-30
Create IIIF manifests to support 3D data


snydman:
Per @blalbrit , the IIIF manifests supporting 3D objects should have the following characteristics. 
- resource with type=3d should get IIIF type "PhysicalObject" not "Document"
- currently, UV doesn't seem to handle any other files as downloadable... could be a type or config error for the manifest, this doesn't seem to be specified anywhere
- this means that the only thing that currently works in the IIIF3 manifest is when only the 3d resource is present

We should support rendering of both 3js and OBJ files.  

There are some cases in which a single DRUID contains multiple 3D models.  Not sure if UV handles this yet, but Ed Silverton is aware.  

---

The desired behavior is to have 3d resource render, and provide download access to the other files

Ben has provided a few models for further analysis:
1) https://dms-data.stanford.edu/data/manifests/v3test/sj957jc8666/manifest1.json - strips out all non-3d resources, works in UV currently
2) https://dms-data.stanford.edu/data/manifests/v3test/sj957jc8666/manifest2.json - all resources present, does not work in UV currently

For comparison, what is currently being produced by our manifest creator: https://purl.stanford.edu/sj957jc8666/iiif3/manifest

blalbrit:
@aeschylus - does it work for any format other than threejs? @snydman 's lab is not using that format in production. That was a specially-baked example for testing purposes.

aeschylus:
It was thrown in here for reference. I am/the access team is still doing analysis. We need to look into updates to UV and vertex to see if any of them support OBJ. There are good paths to supporting whatever file formats are needed using a threeJS viewer (UV or otherwise), but we're still trying to get a handle on the shortest path toward that.

https://gltf-viewer.donmccurdy.com/
https://threejs.org/examples/webgl_loader_obj.html
https://manu.ninja/webgl-3d-model-viewer-using-three-js/

aeschylus:
@blalbrit Can you say more about “IIIF Type physicalObject”. Is this is the Specs somewhere? I can't seem to find it in Presentation 3.0. Is it in an annex?

blalbrit:
@aeschylus - it's absolutely not in the specs. This is coming from whatever Ed specified. See his examples at https://github.com/edsilv/iiif-3d-manifests

blalbrit:
Since IIIF does not have 3D support, this is whatever Ed cooked up for UV.

PR #280, opened 2019-01-09
Generate special IIIF v3 manifests for 3d objects that exclude all re…

cbeer:
…sources except the 3d one so it works in UV.

Part of #269 
Part of sul-dlss/sul-embed#940


cbeer:
Branch deployed to UAT: https://sul-purl-uat.stanford.edu/bg387kw8222

PR #345, opened 2019-08-19
Update service description to v1.0 of Search spec fixes #199

### LINKED ISSUE #199: Content search service description points to wrong version of spec
The content search service description introduced in sul-dlss/purl#198 adds the service description to manifests like the following:

```json
{
  "service": [
    {
      "@context": "http://iiif.io/api/search/0/context.json",
      "@id": "https://exam.pl/cc842mn9348/search",
      "profile": "http://iiif.io/api/search/0/search",
      "label": "Search within this manifest"
    }
  ]
}
```

We should be implementing version 1.0 of the Content Search API, and thus have a service description more like the following:

```json
{
  "service": [
    {
      "@context": "http://iiif.io/api/search/1/context.json",
      "@id": "https://exam.pl/cc842mn9348/search",
      "profile": "http://iiif.io/api/search/1/search",
      "label": "Search within this manifest"
    }
  ]
}
```

See [§3.1 of the spec](http://iiif.io/api/search/1.0/#service-description) for more information.

mejackreed:


cbeer:
As far as I can tell, UV still doesn't support v1 of the spec. I don't know how much of an issue that is for us anymore though?

mejackreed:
@cbeer I don't think that should be a factor for us anymore. I'd propose we 🚢 this

PR #406, opened 2020-06-03
Add ocr_files as seeAlso for canvases

### LINKED ISSUE #405: Provide links to transcriptions in IIIF manifests
By providing our transcriptions in the IIIF manifests, we could enable use in Johannes' Mirador 3 plugin.

mejackreed:
Fixes #405


mejackreed:
Looks like we also need to provide the profile so the plugin will pick it up.

PR #422, opened 2020-07-24
Pass collection information along with the hit

cbeer:


app/assets/javascripts/analytics.js
```diff
@@ -1,52 +1,25 @@
...
   return GoogleAnalytics;
-
 })();
 
 $(document).on("turbolinks:load", function(){
   GoogleAnalytics.load();
-  if (GoogleAnalytics.analyticsId){
-    GoogleAnalytics.trackPageview();
-  }
+  window.ga && window.ga('send', 'pageview', $('link[rel="up"]').attr('href'));
```
mejackreed, line 24: I read this as for every page view, we also add a page view for the collection? Is that correct?

mejackreed: And do we also need to fire a pageview for the current page or is that happening already?

PR #490, opened 2021-05-13
Respond to OPTIONS requests with CORS headers that support content negotiation

cbeer:
Follow-up to https://github.com/sul-dlss/purl/pull/271, which added support for content negotiation, but browers need an explicit invitation to perform it 🤷 .

This should unblock https://github.com/sul-dlss/sul-embed/pull/1256


mjgiarlo (approved):
I'm a little surprised Rails doesn't handle this automagically, but this solution is clean.

PR #496, opened 2021-06-03
Submit the feedback form and render flash messages

### LINKED ISSUE #487: form submission sends two duplicate messages
Recently (in the last week or so), we notice that a submission to the purl feedback form results in two, duplicate messages are received by members of the purl-feedback list. As far as I am aware, nothing has changed on the list management side that would result in duplicate messages.

is it a bug in the form?

mjgiarlo:
Connects to #487

This PR wires up the client-side feedback form such that when a user submits it, one or more flash messages successfully render on the screen. It has three components:

1. Render the flash_messages partial (it was not rendering at all before so the user received no feedback, which I suspect is/was leading to duplicate form submissions);
2. Correct the selector in the client-side feedback code to point at the right div selector for flashes; and
3. Use the `.done()` callback on the AJAX request that submits the form, as `.success()` seems to no longer be defined (at least, I was seeing local errors in the JS console...)

I could not otherwise reproduce this bug, so I am hesitant to say this fixes #487 but maaaaaybe it does if the source is user confusion leading to double submits.

### DIFF
--- a/app/assets/javascripts/feedback_form.js
+++ b/app/assets/javascripts/feedback_form.js
@@ -43,7 +43,7 @@ $(document).on("turbolinks:load", function(){
               url: form.action,
               data: valuesToSubmit,
               type: 'post'
-            }).success(function(response){
+            }).done(function(response){
               if (isSuccess(response)){
                 $($el).collapse('hide');
                 $($form)[0].reset();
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/feedback_form.js
```diff
@@ -68,7 +68,7 @@ $(document).on("turbolinks:load", function(){
 
     function renderFlashMessages(response){
       $.each(response, function(i,val){
-        var flashHtml = "<div class='alert alert-" + val[0] + "'><button type='button' class='close' data-dismiss='alert' aria-hidden='true'>&times;</button>" + val[1] + "</div>";
+        var flashHtml = "<div class='alert alert-" + val[0] + "'>" + val[1] + "<a class='close' data-dismiss='alert' href='#'>&times;</a></div>";
```
mjgiarlo, line 71: This makes the JS flash rendering look identical to the server-side rendering.

babel.config.js
```diff
@@ -47,6 +47,12 @@ module.exports = function(api) {
           loose: true
         }
       ],
+      [
```
mjgiarlo, line 50: This makes webpack less spammy.

PR #498, opened 2021-06-08
Relayout the feedback form for consistency

### LINKED ISSUE #421: Update layout of Feedback form
As part of the effort to standardize our feedback forms, below are recommendations for updating the Purl feedback form. Note that the basic layout updates have been made to Earthworks (sul-dlss/earthworks#691).

Here's the current feedback form:

Screen Shot 2020-07-21 at 3 53 30 PM

---

- [x] A: Add a Bootstrap alert-primary component to the top of the form and move the "Reporting from: current page URL" string to that component.

- [x] B: Right-align the form fields, closer to the field inputs.

- [ ] It doesn't appear that we show a Captcha for non-authenticated users. For the same reasons we display it on other apps, we should add the Captcha when the user is not authenticated, as we do in Earthworks, for example:

Screen Shot 2020-07-21 at 4 00 56 PM

Note there is an existing ticket (#192) in this repo about the feedback form; that could probably be closed after doing the work in this ticket.

mjgiarlo:
Fixes #421

This commit contains a number of changes related to tweaking the layout of the feedback form:

* Requested style changes (see screenshots)
* As part of moving to Bootstrap 5, manage jquery via webpacker which gives us more control over load order (JQuery must be loaded before Bootstrap or BS controls do not function---to wit, the collapsible div is currently broken in deployed environments such that it only collapses when submitted not when cancelled)
* Use recaptcha gem to support ReCAPTCHA use case for unauthenticated users
* Minor JavaScript formatting tweaks
* Remove code from feedback_forms.js that doesn't seem necessary anymore, including the link-to-button functionality
* Fix the styling of BS alerts generated from within javascript (currently, error flashes are unstyled)
* Embangify a private method in FeedbackFormsController (stylistic only)
* Load JQuery before Bootstrap in webpacker config

# Screencast

Peek 2021-06-08 14-33

# Screenshots

## BEFORE

before

## AFTER

after

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,8 +20,6 @@ gem 'sass-rails', '>= 6.0'
 gem 'uglifier', '>= 2.7.2'
 # A gem for simple rails environment specific config
 gem 'config'
-# Use jquery as the JavaScript library
-gem 'jquery-rails'
 
 gem 'turbolinks', '~> 5'
 
@@ -48,6 +46,9 @@ gem 'addressable'
 gem 'tophat'
 gem 'rails-file-icons'
 gem 'sul_styles', '~> 0.6'
+
+# Use recaptcha gem to prevent robots spamming the feedback form
+gem 'recaptcha'
 
 # sul-dlss/osullivan#development has early support for generating IIIF v3 manifests
 gem 'iiif-presentation', github: 'sul-dlss/osullivan', branch: 'development'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -195,10 +195,6 @@ GEM
     iso-639 (0.3.5)
     jbuilder (2.11.2)
       activesupport (>= 5.0.0)
-    jquery-rails (4.4.0)
-      rails-dom-testing (>= 1, < 3)
-      railties (>= 4.2.0)
-      thor (>= 0.14, < 2.0)
     json (2.5.1)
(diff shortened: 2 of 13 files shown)


mjgiarlo:
Have a look, @ggeisler ?

ggeisler (changes requested):
I have one suggestion (though I think it might apply to multiple contexts): 

- The alert message dismissal icon should be right-aligned in the alert box, rather than following immediately after the message text as it appears in the screenshots

See https://getbootstrap.com/docs/5.0/components/alerts/#dismissing for an example if what I'm talking about isn't clear.

Other than that, looks good to me from the UI point of view @mjgiarlo.

mjgiarlo:
@ggeisler :speech_balloon: 
> I have one suggestion (though I think it might apply to multiple contexts):
> 
> * The alert message dismissal icon should be right-aligned in the alert box, rather than following immediately after the message text as it appears in the screenshots

Good catch! I missed the `alert-dismissible` class. Here we go:

Screenshot from 2021-06-08 17-10-17

ggeisler (approved):
👍

cbeer (approved):

camillevilla (approved):
Looks great @mjgiarlo! Thanks for the BS5 notes.

PR #540, opened 2022-01-04
Display multivalued abstract/contents as multiple paragraphs

thatbudakguy:
Joining the abstract field on a double newline causes simple_format
to begin a new paragraph tag. Previously, we concatenated the
abstract field without any delimiter.

See https://github.com/sul-dlss/mods_display/issues/54

example item with multivalued abstract: https://purl.stanford.edu/fr878dp2426

### DIFF
--- a/app/views/purl/_mods_abstract_contents.html.erb
+++ b/app/views/purl/_mods_abstract_contents.html.erb
@@ -12,7 +12,7 @@
   <% if contents.respond_to?(:values) %>
     <%= mods_display_label(contents.label) %>
     <dd class="section-contents">
-      <%= simple_format contents.values.join.gsub('&#10;', "\n").html_safe, {}, sanitize: false %>
+      <%= simple_format contents.values.join("\n\n").gsub('&#10;', "\n").html_safe, {}, sanitize: false %>
     </dd>
   <% end %>
 <% end %>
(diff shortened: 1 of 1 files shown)


app/views/purl/_mods_abstract_contents.html.erb
```diff
@@ -3,7 +3,7 @@
   <% if abstracts.respond_to?(:values) %>
     <%= mods_display_label(abstracts.label) %>
     <dd class="section-abstract">
-      <%= simple_format abstracts.values.join.gsub('&#10;', "\n").html_safe, {}, sanitize: false %>
+      <%= simple_format abstracts.values.join("\n\n").gsub('&#10;', "\n").html_safe, {}, sanitize: false %>
```
cbeer, line 6: WFM. We should probably do the same to the contents display below for consistency?

thatbudakguy: It looks like the [LC MODS guidelines](https://www.loc.gov/standards/mods/userguide/tableofcontents.html) say it's "not recommended" to have contents be multivalued, but I guess we support it, so might as well? cc @arcadiafalcone

arcadiafalcone: Yes. It can also happen that the element appears twice with the same info in different languages.

PR #560, opened 2022-03-02
Remove URL-encoded slashes in IIIF manifest URLs

### LINKED ISSUE #559: url-encoded slash in manifest links breaks IIIF spec
this issue is via @mapninja and related to supporting manifest import in the new [allmaps viewer](https://viewer.allmaps.org/).

our presentation v2 manifests currently url-encodes a slash (`%2F`) in between the DRUID and filename of stacks URLs:
https://github.com/sul-dlss/purl/blob/835363d0dfc3cf05ba5ec26560ec4eb3903b72c4/app/models/iiif_presentation_manifest.rb#L360

...resulting in URLs like: `https://stacks.stanford.edu/image/iiif/wy534zh7137%2FSULAIR_rosette`. this URL does resolve to the correct path `https://stacks.stanford.edu/image/iiif/wy534zh7137/SULAIR_rosette`, but it results in manifests listing an `@id` using two different forms for the same image (see for example https://purl.stanford.edu/tc010rc2373/iiif/manifest).

the [prezi v2 spec](https://iiif.io/api/presentation/2.1/#requests) on the "prefix" part of the URL scheme says:
> The prefix may contain multiple path segments, delimited by slashes, but all other special characters must be encoded.

not positive this is the right part of the spec, but it seems like encoding the `/` specifically is discouraged. this makes sense, since opening URLs like this in a browser causes the `%2F` to become double-url-encoded as `%252F`...and you can keep doing this forever to make longer and longer URLs!

the `%2F` is also hardcoded in this location:
https://github.com/sul-dlss/purl/blob/835363d0dfc3cf05ba5ec26560ec4eb3903b72c4/app/models/iiif_presentation_manifest.rb#L108-L109

thatbudakguy:
Fixes sul-dlss/stacks#864

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -14,9 +14,9 @@ class Iiif3PresentationManifest < IiifPresentationManifest
       'label' => { en: [title] },
       'requiredStatement' => iiif_key_value('Attribution', attribution),
       'logo' => [{
-        'id' => 'https://stacks.stanford.edu/image/iiif/wy534zh7137%2FSULAIR_rosette/full/400,/0/default.jpg',
+        'id' => 'https://stacks.stanford.edu/image/iiif/wy534zh7137/SULAIR_rosette/full/400,/0/default.jpg',
         'type' => 'Image',
-        'service' => [iiif_image_v2_service('https://stacks.stanford.edu/image/iiif/wy534zh7137%2FSULAIR_rosette')]
+        'service' => [iiif_image_v2_service('https://stacks.stanford.edu/image/iiif/wy534zh7137/SULAIR_rosette')]
       }],
       'seeAlso' => [{
         'id' => controller.purl_url(druid, format: 'mods'),
--- a/app/models/iiif_presentation_manifest.rb
+++ b/app/models/iiif_presentation_manifest.rb
@@ -105,8 +105,8 @@ class IiifPresentationManifest
       'label' => title,
       'attribution' => copyright || 'Provided by the Stanford University Libraries',
       'logo' => {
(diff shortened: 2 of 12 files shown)


cbeer (changes requested):
I'm tossing "request changes" on this just to make sure we do some due diligence with any IIIF annotations from e.g. Parker to make sure we don't break any linkages.

anarchivist:
I'm also not certain that this is the solution - the fix is probably percent-encoding the slashes in the URLs in the image's info.json responses rather than than not-encoding them in the manifest (see https://iiif.io/api/image/2.1/#identifier).

thatbudakguy:
@anarchivist thanks for digging that up; I suspected I just hadn't found the right spec to read. will make a new PR approaching this from the opposite direction

cbeer:
I'm reopening this for our further consideration. If we stop encoding slashes, we can stop fighting against Rails, browsers, etc. We're ok spec-wise if we just imagine this as shuffling the initial druid into to the IIIF service prefix, e.g.:

IIIF Image URL:
{scheme}://{server}{/prefix}/{identifier}

https://stacks.stanford.edu/image/iiif/vg994wz9415%252Fvg994wz9415_0001/info.json
server: stacks.stanford.edu
prefix: /image/iiif/
identifier: vg994wz9415%252Fvg994wz9415_0001

[https://stacks.stanford.edu/image/iiif/vg994wz9415/vg994wz9415_0001/info.json](https://stacks.stanford.edu/image/iiif/vg994wz9415%252Fvg994wz9415_0001/info.json)
server: stacks.stanford.edu
prefix: /image/iiif/vg994wz9415
identifier: vg994wz9415_0001

This is probably preferable than trying to change things on stacks to avoid anything breaking (e.g. clients relying on particular encodings or pre-existing annotations). That said, we make no claims about the persistence or durability of stacks links, and our self-generated annotations all seem to be (appropriately) canvas-level... so maybe rejecting the encoded format is something to consider for the future.

PR #640, opened 2023-01-10
Don't initialize the javascript more than one time

### LINKED ISSUE #639: Two embed displaying
This was a regression *after* ~~https://github.com/sul-dlss/purl/commit/1491f47834f9507b421e6e9443cec888a61e6edc~~ 43a8cc17df7d3bd50f6ec127cfd37bfe94cd9d11

This event is triggering twice https://github.com/sul-dlss/purl/blob/master/app/javascript/purl_embed.js#L4

jcoyne:
Fixes #639


marlo-longley:
Nice !

PR #667, opened 2023-03-21
Use stimulus oembed

jcoyne:
This is the same script we're using with arclight.  This no longer requires jquery.


config/importmap.rb
```diff
@@ -4,6 +4,9 @@
 pin "jquery", to: "https://ga.jspm.io/npm:jquery@3.6.1/dist/jquery.js"
 pin "bootstrap", to: "https://ga.jspm.io/npm:bootstrap@5.2.3/dist/js/bootstrap.esm.js"
 pin "@popperjs/core", to: "https://ga.jspm.io/npm:@popperjs/core@2.11.6/lib/index.js"
+pin "@hotwired/stimulus", to: "stimulus.min.js", preload: true
+pin "@hotwired/stimulus-loading", to: "stimulus-loading.js", preload: true
+pin_all_from "app/javascript/controllers", under: "controllers"
 pin 'jquery.oembed'
```
thatbudakguy, line 10: probably don't need jquery.oembed here anymore either, right?

thatbudakguy (approved):

config/importmap.rb
jcoyne: Thank you. I removed that.

PR #697, opened 2023-07-24
Generate the link to searchworks using the FOLIO hrid value if there's no catkey

cbeer:


andrewjbtw:
If I'm understanding this correctly, if we remove the catkey from the SDR metadata post-migration, then the purl link to SW will break because there's only a Folio instance HRID, and that will have the prefixed value. Is that correct? 

Post-migration, we won't be adding new catkey values, only Folio instance HRIDs, so those would also be prefixed values. That also seems like a potential problem.

andrewjbtw:
Ah - thanks. So we can publish either one and be ok. We haven't yet made a decision on removing the catkey field but I think we will eventually do that.

app/models/public_xml.rb
```diff
@@ -18,10 +18,11 @@ def content_metadata
   end
 
   def catalog_key
+    @catalog_key ||= document.root.at_xpath('identityMetadata/otherId[@name="catkey"]').try(:text).presence
     @catalog_key ||= begin
-      key = document.root.at_xpath('identityMetadata/otherId[@name="catkey"]').try(:text)
-
-      key if key.present?
+      key = document.root.at_xpath('identityMetadata/otherId[@name="folio_instance_hrid"]').try(:text).presence
+      key = key.delete_prefix('a') if key&.match?(/^a\d+$/)
```
jcoyne, line 24: Is there a reason we need to see if it matches this pattern?  Won't all folio_instance_hrids that begin with an "a" need to have it removed?

In other words, can we just do `key = key&.delete_prefix('a')`?

cbeer: I guess I'm not sure how arbitrary FOLIO hrids are; we've been pretty cagey about how we're handling them in [Searchworks](https://github.com/sul-dlss/SearchWorks/pull/3296/files#diff-959bc9abc46a55332bb64d5155a79323afa75a50ec1a2137ddd22d926f62c6c5R5) and [requests](https://github.com/sul-dlss/sul-requests/blob/main/app/models/folio/instance.rb#L93).

jcoyne (approved):

PR #724, opened 2023-10-23
Adds items in this collection section.

### LINKED ISSUE #714: Collection PURL pages should support exploring the collection
### proposal
[view proposal in requirements doc](https://docs.google.com/document/d/1W5rmHTMRGso5uxqIKqptvUYVYIn9cUeo6Nm9rx96XTU/edit#heading=h.oh14q4bdbcbl)

Provide a mechanism from the collection PURL page to find out what items are in the collection. Link text should make it clear what it does and should be in an easily findable location on the page.

Currently, the plan is to do this by linking to SearchWorks using the `collection` parameter, e.g.  https://searchworks.stanford.edu/catalog?f%5Bcollection%5D%5B%5D=cj269gn0736

### user story
_As a user who has identified a collection PURL of interest, I would like to be able to easily find out more about the items in the collection._

### design
Screenshot 2023-10-18 at 11 08 58

### questions
Need to account for cases where the collection object has not been released. It seems like content in these collections is somewhat unrelated to each other so there would not need to be links on the item PURLs to a collection or the other items in it. See the requirements doc for detailed info.

### out of scope
We don't need to actually list any information about the collection members on the collection purl page.

justinlittman:
closes #714

image

### DIFF
--- a/app/models/public_xml.rb
+++ b/app/models/public_xml.rb
@@ -45,4 +45,8 @@ class PublicXml
       node.attribute('resource').text.split('/', 2).last.split(':', 2).last
     end
   end
+
+  def object_type
+    document.root.at_xpath('identityMetadata/objectType').try(:text)
+  end
 end
--- a/app/views/purl/_collection_items.html.erb
+++ b/app/views/purl/_collection_items.html.erb
@@ -1,0 +1,3 @@
+<% if @purl.object_type == 'collection' && @purl.catalog_key %>
+  <%= link_to 'View items in this collection in SearchWorks', "#{Settings.searchworks.url}/catalog?f[collection][]=#{@purl.catalog_key}" %>
+<% end %>
--- a/app/views/purl/_mods_metadata_sections.html.erb
+++ b/app/views/purl/_mods_metadata_sections.html.erb
@@ -7,6 +7,7 @@
   bibliography = render(partial: "mods_bibliographic", locals: {document: document})
   contact = render(partial: "mods_contact", locals: {document: document})
   find_it = render(partial: "find_it", locals: {document: document})
+  collection_items = render(partial: "collection_items", locals: {document: document})
 %>
 
 <% if description.present? %>
@@ -84,6 +85,18 @@
     </div>
   </div>
 <% end %>
+
+<% if collection_items.present? %>
+  <div class="section" id="<%= t('record_side_nav.find_it.id') %>" data-side-nav-class="<%= t('record_side_nav.find_it.id') %>">
+    <div class="section-heading">
+      <h2><%= t('record_side_nav.find_it.icon').html_safe %> Items in collection</h2>
+    </div>
+    <div class="section-body">
+      <%= collection_items %>
+    </div>
+  </div>
+<% end %>
+
 
 <% if find_it.present? %>
   <div class="section" id="<%= t('record_side_nav.find_it.id') %>" data-side-nav-class="<%= t('record_side_nav.find_it.id') %>">
--- a/spec/model/purl_resource_spec.rb
+++ b/spec/model/purl_resource_spec.rb
@@ -291,4 +291,20 @@ RSpec.describe PurlResource do
       end
     end
   end
+
+  describe '#object_type' do
+    it 'pulls the value from the identity metadata' do
+      allow(subject).to receive(:public_xml_body).and_return <<-EOF
+      <?xml version="1.0" encoding="UTF-8"?>
+      <publicObject>
+        <identityMetadata>
+          <objectType>collection</objectType>
+          <objectLabel>Acquisitions Serials</objectLabel>
+        </identityMetadata>
+      </publicObject>
+      EOF
+
+      expect(subject.object_type).to eq 'collection'
+    end
+  end
 end
--- a/spec/views/purl/_collection_items.html.erb_spec.rb
+++ b/spec/views/purl/_collection_items.html.erb_spec.rb
@@ -1,0 +1,35 @@
+require 'rails_helper'
+
+RSpec.describe 'purl/_collection_items' do
+  before do
+    assign(:purl, purl)
+  end
+
+  context 'when collection has a catalog record id' do
+    let(:purl) { PurlResource.new(id: 'bb631ry3167') }
+
+    it 'displays a View items in this collection link' do
+      render
+      expect(rendered).to have_css 'a[href="https://searchworks.stanford.edu/catalog?f[collection][]=13965062"]',
+                                   text: 'View items in this collection in SearchWorks'
+    end
+  end
+
+  context 'when collection does not have a catalog record id' do
+    let(:purl) { PurlResource.new(id: 'gk894yk3598') }
+
+    it 'does not display View items in this collection link' do
+      render
+      expect(rendered).not_to have_css 'a'
+    end
+  end
+
+  context 'when not a collection' do
+    let(:purl) { PurlResource.new(id: 'cg357zz0321') }
+
+    it 'does not display View items in this collection link' do
+      render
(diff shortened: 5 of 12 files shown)


app/models/public_xml.rb
```diff
@@ -45,4 +45,8 @@ def relations(predicate)
       node.attribute('resource').text.split('/', 2).last.split(':', 2).last
     end
   end
+
+  def object_type
+    document.root.at_xpath('identityMetadata/objectType').try(:text)
```
jcoyne, line 50: Why try here? There shouldn't be any nodes with out an objectType, right? That's pretty fundamental.

app/views/purl/_collection_items.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if @purl.object_type == 'collection' && @purl.catalog_key %>
```
jcoyne, line 1: Can you make a method `collection?` here?

app/views/purl/_collection_items.html.erb
```diff
@@ -0,0 +1,3 @@
...
+  <%= link_to 'View items in this collection in SearchWorks', "#{Settings.searchworks.url}/catalog?f[collection][]=#{@purl.catalog_key}" %>
```
jcoyne, line 2: Is it possible the collection is not released to searchworks?

spec/views/purl/_collection_items.html.erb_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+RSpec.describe 'purl/_collection_items' do
+  before do
+    assign(:purl, purl)
+  end
+
+  context 'when a published collection' do
+    let(:purl) { PurlResource.new(id: 'bb631ry3167') }
+
+    it 'displays a View items in this collection link' do
```
jcoyne, line 11: Isn't this same scenario tested in the integration test?

spec/views/purl/_collection_items.html.erb_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+                                   text: 'View items in this collection in SearchWorks'
+    end
+  end
+
+  context 'when an unpublished collection' do
+    let(:purl) { PurlResource.new(id: 'gk894yk3598') }
+
+    it 'does not display View items in this collection link' do
+      render
+      expect(rendered).not_to have_css 'a', text: 'View items in this collection in SearchWorks'
```
jcoyne, line 23: These two negative assertion tests are very fragile because a change to the text in the template that does not also change these tests completely makes these assertions useless. E.g we could show a link, but since the text has changed we won't catch it.

app/views/purl/_mods_metadata_sections.html.erb
```diff
@@ -85,6 +86,18 @@
   </div>
 <% end %>
 
+<% if collection_items.present? %>
```
jcoyne, line 89: Can't you just move all of this into the partial?  That is already wrapped in a conditional.  Alternatively if you introduce a ViewComponent here, you can make this be the `render?` condition.  I find that much easier to read and test rather than this approach here.

spec/model/purl_resource_spec.rb
```diff
@@ -291,4 +291,20 @@
       end
     end
   end
+
+  describe '#object_type' do
+    it 'pulls the value from the identity metadata' do
+      allow(subject).to receive(:public_xml_body).and_return <<-EOF
```
jcoyne, line 297: Please move this to a `before` as it's part of setting up the scenario and has no assertions.

jcoyne (changes requested):

app/views/purl/_collection_items.html.erb
jcoyne: Nope, because it has a catalog key.

spec/views/purl/_collection_items.html.erb_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    let(:purl) { PurlResource.new(id: 'bb631ry3167') }
+
+    it 'displays a View items in this collection link' do
+      render
+      expect(rendered).to have_css 'a[href="https://searchworks.stanford.edu/catalog?f[collection][]=13965062"]',
+                                   text: 'View items in this collection in SearchWorks'
+    end
+  end
+
+  context 'when an unpublished collection' do
```
jcoyne, line 18: How do we know it's "unpublished"?  Do you mean "doesn't have a catkey"?  I'm concerned that we're overloading what "published" means.

app/models/public_xml.rb
justinlittman: Because every other call to `.text` in this class is `.try(:text)`. I have no idea why that would be, but following existing conventions.

app/views/purl/_collection_items.html.erb
justinlittman: I would have, except `PurlResource` already has a `.collection` method which provides the collection that the item is a part of. It seems that a `.collection?` method that indicates whether this item is a collection risks confusion.

spec/views/purl/_collection_items.html.erb_spec.rb
justinlittman: Yes, but I wanted the test cases in this spec to be complete so that a dev could understand the expected behavior from the view from this alone.

The integration test is necessary because it tests that the view is actually called.

spec/views/purl/_collection_items.html.erb_spec.rb
justinlittman: I couldn't figure out how to assert nothing is rendered, so just removed the text assertion.

app/views/purl/_mods_metadata_sections.html.erb
justinlittman: I agree with you, but I am following the existing pattern in this file. I'm not trying to refactor in this PR.

spec/model/purl_resource_spec.rb
justinlittman: I'm just following the pattern in this file.

spec/views/purl/_collection_items.html.erb_spec.rb
justinlittman: OK, changed to `when collection does not have a catalog record id`.

app/views/purl/_collection_items.html.erb
jcoyne: I still feel that we should move as much logic from the views as possible.  If we need to rename `collection` to `is_member_of`, I think that would be good to and help align with Cocina terminology.

app/models/public_xml.rb
jcoyne: I don't think that's a very good reason. A lot of this code was written 12 years ago before the model was clear and before there was a safe navigation operator in Ruby.

spec/model/purl_resource_spec.rb
jcoyne: This codebase is full of bad patterns and needs to be over hauled. Please use practices like https://www.betterspecs.org/#mock and not the existing patterns here.

app/views/purl/_collection_items.html.erb
jcoyne: This method name has been refactored, so please move this to the `collection?` method now.

app/views/purl/_mods_metadata_sections.html.erb
jcoyne: I've changed the existing pattern in https://github.com/sul-dlss/purl/pull/727 so please follow that pattern here.

jcoyne:
There are now conflicts with the DOI feature.

justinlittman:
Superceded by #736

PR #725, opened 2023-10-24
Rename collection to containing_collection

jcoyne:
This helps be clearer that we are refering to the collection that contains this item and not the items this collection contains or whether or not this is a collection type of resource


ndushay (approved):
oh no!  conflicts!

Issue #718, opened 2023-10-18
Display Altmetric badge on PURL pages


thatbudakguy:
### proposal
[see proposal in requirements doc](https://docs.google.com/document/d/1W5rmHTMRGso5uxqIKqptvUYVYIn9cUeo6Nm9rx96XTU/edit#heading=h.7d870vk45abw)

- [ ] Modify PURL markup so that it includes the [elements required](https://help.altmetric.com/support/solutions/articles/6000240582-required-metadata-for-content-tracking) for SDR deposits to be tracked by Altmetric. 
  - This must include two of the supported meta tags on Altmetric's list, one of which must be an identifier on [their list](https://help.altmetric.com/support/solutions/articles/6000240585-scholarly-identifiers-supported-by-altmetric) (DOIs are on the list, PURLs are not).
  - They recommend as many of the supported meta tags as possible.
- [ ] Display of Altmetric badges on PURL pages in compliance with [Altmetric guidelines](https://www.altmetric.com/solutions/free-tools/institutional-repository-badges/).
  - [Technical documentation](https://badge-docs.altmetric.com/)
  - Note that since PURL is not one of the identifiers they track, we will only be able to track PURLs that also have a DOI, and tracking will only include mentions of the DOI and not the PURL (in cases where someone used the PURL but not the DOI for a citation).

### user stories
_As a depositor, I would like to see Altmetric badges displayed on my content so that I can better understand the broader impacts of my work and demonstrate this to funders, promotion and tenure committees, potential employers, etc._

_As a university administrator, I would like to see Altmetric information about content from my faculty and other department members so that I can better understand and promote the work of my department, research institute, or school._

### design
- we're using the smallest "donut" design in the options that Altmetric provides.
- the popover on hover should be enabled.
- other details should not be enabled.

![Screenshot 2023-10-18 at 12 32 49](https://github.com/sul-dlss/purl/assets/4924494/428f06a5-487b-4a18-afc3-0ff14cfc639a)

[see it in context in the invision mockup](https://projects.invisionapp.com/share/FV13CCPQXMHA#/screens/474686424)

lwrubel:
@amyehodge will provide text for every altmetrics-eligible page.

amyehodge:
Text that should appear always when an item is being tracked by Altmetric is 
"Altmetric tracks online attention to this work."
[...]

PR #728, opened 2023-10-24
Adds citation_doi meta for altmetrics.

justinlittman:
refs #718

image

### DIFF
--- a/app/models/purl_resource.rb
+++ b/app/models/purl_resource.rb
@@ -182,6 +182,16 @@ class PurlResource
     def representative_thumbnail
       "#{iiif_manifest.thumbnail_base_uri}/full/!400,400/0/default.jpg" if iiif_manifest.thumbnail_base_uri.present?
     end
+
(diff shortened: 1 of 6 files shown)


thatbudakguy (approved):

amyehodge:
This needs at least title markup as well. From Altmetric:

"We can see the identifier metatags have been included and we are able successfully to pick these up and read them, though to begin displaying entries for the outputs we are still short a few bits of metadata mentioned in the [Required metadata for content tracking](https://help.altmetric.com/support/solutions/articles/6000240582-required-metadata-for-content-tracking) details sent over, in particular there doesn't appear to be a trackable Title tag, which would be required to populate a badge/details page. We would also recommend adding an author tags as well so these can be displayed on the details page. 
​
We'd recommend using the <citation_title> <citation_author> tags but there are also some further alternatives on the link above, once these are in place we can get all the data we need to create the details pages for the outputs."

PR #732, opened 2023-10-24
Display link to item's collection's item.

### LINKED ISSUE #715: Collection member PURL pages should support exploring the collection
Same as #714, but from the perspective of the collection member object.

### proposal
[view proposal in requirements doc](https://docs.google.com/document/d/1W5rmHTMRGso5uxqIKqptvUYVYIn9cUeo6Nm9rx96XTU/edit#heading=h.oh14q4bdbcbl)

Provide a mechanism from the item PURL page to access the collection and find out what other items it contains. Link text should make it clear what it does and should be in an easily findable location on the page.

Currently, the plan is to do this by linking to SearchWorks using the `collection` parameter, e.g.  https://searchworks.stanford.edu/catalog?f%5Bcollection%5D%5B%5D=cj269gn0736

### user story
_As a user who has identified an item PURL of interest, I would like to be able to easily find more information about the collection this item is part of and what other items are in this collection._

### design
Screenshot 2023-10-18 at 11 03 31

See [the full design on invision](https://projects.invisionapp.com/share/FV13CCPQXMHA#/screens/474686424).

### questions
Need to account for cases where the collection object has not been released. It seems like content in these collections is somewhat unrelated to each other so there would not need to be links on the item PURLs to a collection or the other items in it. See the requirements doc for detailed info.

justinlittman:
closes #715

image

### DIFF
--- a/app/views/purl/_collection.html.erb
+++ b/app/views/purl/_collection.html.erb
@@ -1,0 +1,15 @@
+<% if document.containing_purl_collections.present? %>
+  <div class="section">
+    <div class="section-heading">
+      <h2>Collection</h2>
+    </div>
+    <div class="section-body">
+      <% document.containing_purl_collections.each do |collection_document| %>
+        <p class="mb-2"><%= collection_document.title %></p>
+        <% if collection_document.folio_instance_hrid %>
+          <%= link_to('View other items in this collection in SearchWorks', "#{Settings.searchworks.url}/catalog?f[collection][]=#{collection_document.folio_instance_hrid}") %>
+        <% end %>
+      <% end %>
+    </div>
+  </div>
+<% end %>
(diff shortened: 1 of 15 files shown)


app/views/purl/_collection.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if document.containing_purl_collections.any? %>
```
jcoyne, line 1: Should we just use `.present?` here instead of `.any?` We don't need to test the truthiness of any of these values, right?

app/views/purl/show.html.erb
```diff
@@ -6,7 +6,7 @@
   <% end %>
   <%= tag :link, rel: "alternate", title: "MODS XML", type: 'application/xml', href: purl_url(@purl, format: 'mods') if @purl.mods? %>
   <%= tag :link, rel: "alternate", title: "IIIF Manifest", type: 'application/json', href: iiif_manifest_url(@purl) if @purl.iiif_manifest? %>
-  <%= tag :link, rel: "up", href: purl_url(@purl.containing_collection) if @purl.containing_collection.present? %>
+  <%= tag :link, rel: "up", href: purl_url(@purl.containing_collections.first) if @purl.containing_collections.any? %>
```
jcoyne, line 9: Again, I think `present?` is ideal here unless we're trying to filter out falsey values.

jcoyne (changes requested):

jcoyne (approved):

PR #737, opened 2023-10-25
Extract a PreviewResource for displaying the preview

jcoyne:
Because meta-programming is harder to reason about than having regular methods

### DIFF
--- a/app/controllers/preview_controller.rb
+++ b/app/controllers/preview_controller.rb
@@ -2,17 +2,8 @@ class PreviewController < ApplicationController
   def index; end
 
   def show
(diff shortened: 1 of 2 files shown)


justinlittman:
Do you know if the MODS preview still used? If so, where and why?

jcoyne:
@justinlittman Arcadia said:
>I've used it in the past to test how metadata should be structured to achieve a particular result in MODS display, and how display quirks might be rectified.

justinlittman (approved):
Approve pending green.

jcoyne:
@justinlittman I'm keeping all the green in https://github.com/sul-dlss/purl/pull/738

PR #743, opened 2023-10-26
Set up analytics tracking with ahoy

### LINKED ISSUE #720: Install and configure ahoy for metrics collection
### proposal
[Ahoy](https://github.com/ankane/ahoy) provides a simple first-party analytics solution that gives us more control than google analytics.

- The implementation of metrics collection for PURLs using ahoy has two pieces: this is the server-side piece.
- The work to actually collect metrics will happen on the client side in sul-embed, using [ahoy.js](https://github.com/ankane/ahoy).
- Ahoy sets up a new API in PURL to which embedded apps can report analytics.
- Once it is installed, events (in the form of `Event` objects) are saved to a database (note that Purl doesn't currently have a db).

view the [requirements analysis of ahoy](https://docs.google.com/document/d/1W5rmHTMRGso5uxqIKqptvUYVYIn9cUeo6Nm9rx96XTU/edit#heading=h.noqmyhw0ebos) in the requirements doc.

### steps

- [x] install the ahoy gem
- [x] kick the tires on the API, maybe using `curl`, to see if events can be created and tied to a `PurlResource`
- [x] make sure ahoy does the right thing in local dev vs production-like environments

thatbudakguy:
Closes #720 

- Set up Ahoy for analytics tracking
- Set up a SQLite database for tracking analytics
- Add view and download tracking accessors to PurlResource

See the changes to the README for an explanation of how to fake analytics events in local development.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,22 +1,25 @@
 # SQLite version 3.x
-#   gem install sqlite3-ruby (not necessary on OS X Leopard)
-development:
+#   gem install sqlite3
+#
+#   Ensure the SQLite 3 gem is defined in your Gemfile
+#   gem 'sqlite3'
+#
+default: &default
   adapter: sqlite3
-  database: db/development.sqlite3
   pool: 5
   timeout: 5000
+
+development:
+  <<: *default
+  database: db/development.sqlite3
 
 # Warning: The database defined as "test" will be erased and
 # re-generated from your development database when you run "rake".
 # Do not set this db to the same as development or production.
 test:
-  adapter: sqlite3
+  <<: *default
   database: db/test.sqlite3
-  pool: 5
-  timeout: 5000
 
 production:
-  adapter: sqlite3
+  <<: *default
   database: db/production.sqlite3
-  pool: 5
-  timeout: 5000
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -40,6 +40,27 @@ RSpec.configure do |config|
   # instead of true.
   config.use_transactional_fixtures = false
 
+  # Clean up the database for each test
+  config.before(:suite) do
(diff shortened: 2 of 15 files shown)


spec/rails_helper.rb
```diff
@@ -40,6 +40,27 @@
   # instead of true.
   config.use_transactional_fixtures = false
```
jcoyne, line 41: Can't you just set this to `true` and then you don't need `DatabaseCleaner`?

jcoyne (changes requested):

config/database.yml
```diff
@@ -1,22 +1,25 @@
...
 test:
-  adapter: sqlite3
+  <<: *default
   database: db/test.sqlite3
-  pool: 5
-  timeout: 5000
 
 production:
-  adapter: sqlite3
+  <<: *default
```
jcoyne, line 24: Purl prod is load balanced, so you aren't going to be able to use a sqlite database, right?

jcoyne (changes requested):

config/database.yml
jcoyne: I see #744

thatbudakguy: Yeah, this defers the solution for prod for the time being.

spec/rails_helper.rb
thatbudakguy: I tried that and it didn't actually clean the database. 🤷🏻 I copied this config from another app where we used this setup, and it did what I wanted.

jcoyne: There is almost no reason for it in a modern Rails app.  Please don't stuff this in this PR.

thatbudakguy:
superseded by https://github.com/sul-dlss/metrics/pull/6

PR #749, opened 2023-10-30
Re-add the above-the-fold collection display

thatbudakguy:
This was removed in 6fc28fa.

### DIFF
--- a/app/views/purl/_mods_upper_metadata_section.html.erb
+++ b/app/views/purl/_mods_upper_metadata_section.html.erb
@@ -1,10 +1,11 @@
 <% document_notes = document.mods.note.select { |x| x.label =~ /Preferred citation/i } %>
-<% if document_notes.present? %>
+<% fields = document_notes + document.mods.collection %>
+<% if fields.any? %>
   <div class="section">
     <div class="section-body">
       <dl>
-        <% document_notes.each do |notes| %>
-          <%= mods_record_field(notes) %>
+        <% fields.each do |field| %>
+          <%= mods_record_field(field) %>
         <% end %>
       </dl>
     </div>


app/views/purl/_mods_upper_metadata_section.html.erb
```diff
@@ -1,11 +1,14 @@
 <% document_notes = document.mods.note.select { |x| x.label =~ /Preferred citation/i } %>
-<% if document_notes.present? %>
+<% if document_notes.present? || document.mods.collection.any? %>
```
jcoyne, line 2: Can we combine `fields = document_notes + document.mods.collection` and iterate over that. Both are handled the same way.

jcoyne:
@thatbudakguy won't this display the collections twice on the page?

thatbudakguy:
yeah, it will. I think we're waiting to hear back about whether that was an intentional choice in the design (which I could see it being).

astridu:
@amyehodge Also commented in Slack but will also put this comment in the ticket. It was intentional because we were going to have the collection in the right column, but since we're not doing that, not sure if Amy wants to keep the collection name under the the item title?

amyehodge:
Since we are not currently doing the major redesign with the two columns, everything on the page should stay where it has been, so the collection name should be up above the viewer. @astridu does that mean the new link should be there as well? Or is that going further down the page?

astridu:
@amyehodge The link will be under the Collections section in the body of the page. I put the collections line back at the top.

app/views/purl/_mods_upper_metadata_section.html.erb
thatbudakguy: updated

jcoyne (approved):

PR #752, opened 2023-10-31
Adds navPlace to iiif manifest.

### LINKED ISSUE #708: Add navPlace property to IIIF manifests for objects with defined coordinates
The IIIF presentation API has finalized its [`navPlace` extension](https://iiif.io/api/extension/navplace/), which allows manifests to describe the geographic extent of objects using geoJSON-LD. There are already many client applications beginning to leverage this functionality for rich presentation of content:

- https://mikeapp.github.io/maptime-demo/
- https://centerfordigitalhumanities.github.io/navplace-viewer/?iiif-content=https://centerfordigitalhumanities.github.io/navplace-viewer/tests/fixtures/collection-2.json
- https://eliotjordan.github.io/navplace/

Some scanned maps in the SDR have coordinate information already stored as part of the record. If this information exists and is in lat/long format, we could convert it to geoJSON and serve it as part of the manifest so that the map could be displayed in its geographic context.

PURL generates IIIF manifests (specifically, v3) using the `Iiif3PresentationManifest` model. This issue involves changes to v3, but not v2. I think we still render and send v2 because v3 doesn't include all the properties that some sul-embed viewers need (?); see https://github.com/sul-dlss/purl/issues/164. You can request v3 specifically via the `/[druid]/iiif3/manifest` endpoint of PURL.

justinlittman:
closes #708

A point:
image

A polygon:
image

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -1,8 +1,10 @@
 require 'iiif/presentation'
 require 'iiif/v3/presentation'
+require 'geo/coord'
 
 class Iiif3PresentationManifest < IiifPresentationManifest
   delegate :reading_order, to: :content_metadata
+  attr_reader :purl_base_uri
 
   # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
   def body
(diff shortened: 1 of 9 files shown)


app/models/iiif3_presentation_manifest.rb
```diff
@@ -282,4 +288,97 @@ def iiif_location_auth_service
   def three_d?
     type == '3d'
   end
+
+  def nav_place
+    @nav_place ||= NavPlace.new(public_xml_document:, purl_base_uri:).build
+  end
+
+  class NavPlace
```
jcoyne, line 296: Can you decouple this class from `public_xml_document`? Then would you please move this into https://github.com/iiif-prezi/osullivan ?  That is where we keep all the manifest building tools.

jcoyne (changes requested):

thatbudakguy:
after some hands-on testing today, I'm going to merge this, noting that #761 should be the next step.

thatbudakguy (approved):

Issue #757, opened 2023-10-31
Improve purl performance


justinlittman:
![image](https://github.com/sul-dlss/purl/assets/588335/d55ff147-d72b-4a77-b8ec-674beeba0485)

![image](https://github.com/sul-dlss/purl/assets/588335/39df1abd-515a-4a43-88e3-620361f35492)

```
13:50:11 web.1  | Started GET "/rp193xx6845" for 127.0.0.1 at 2023-10-31 13:50:11 -0400
13:50:11 web.1  | Processing by PurlController#show as HTML
13:50:11 web.1  |   Parameters: {"id"=>"rp193xx6845"}
13:50:11 web.1  | Fetching rp193xx6845 public_xml at https://purl.stanford.edu/rp193xx6845.xml (336.7ms)
13:50:11 web.1  |   Rendering layout layouts/application.html.erb
13:50:11 web.1  |   Rendering purl/show.html.erb within layouts/application
13:50:12 web.1  | Fetching rp193xx6845 mods at https://purl.stanford.edu/rp193xx6845.mods (332.9ms)
13:50:12 web.1  |   Rendered purl/_mods_upper_metadata_section.html.erb (Duration: 1.4ms | Allocations: 1139)
13:50:12 web.1  |   Rendered purl/_embed.html.erb (Duration: 0.9ms | Allocations: 797)
13:50:12 web.1  |   Rendered purl/_access_metadata.html.erb (Duration: 24.3ms | Allocations: 11773)
13:50:12 web.1  |   Rendered purl/_access.html.erb (Duration: 25.3ms | Allocations: 12272)
13:50:12 web.1  |   Rendered purl/_mods_description.html.erb (Duration: 140.0ms | Allocations: 68208)
13:50:12 web.1  |   Rendered purl/_mods_contributors.html.erb (Duration: 132.0ms | Allocations: 39383)
13:50:12 web.1  |   Rendered purl/_mods_abstract_contents.html.erb (Duration: 4.6ms | Allocations: 1365)
13:50:12 web.1  |   Rendered purl/_mods_subjects.html.erb (Duration: 13.5ms | Allocations: 12321)
13:50:12 web.1  |   Rendered purl/_mods_bibliographic.html.erb (Duration: 563.8ms | Allocations: 310092)
13:50:12 web.1  |   Rendered purl/_mods_contact.html.erb (Duration: 0.7ms | Allocations: 368)
[...]

PR #759, opened 2023-10-31
Switch feedback form to turbo.

justinlittman:
refs #757

Goal is to address this:
image


thatbudakguy (changes requested):
I think we need to tweak some markup or CSS. This results in the form being shifted to the left:

Screenshot 2023-10-31 at 13 48 34

Vs. production:

Screenshot 2023-10-31 at 13 48 11

Tested on chromium v118.

justinlittman:
Thanks for the catch @thatbudakguy. Fixed:
image

thatbudakguy (approved):

Issue #721, opened 2023-10-18
Display metrics on PURL pages


thatbudakguy:
### proposal
requires #720.

[view this issue in the requirements doc](https://docs.google.com/document/d/1W5rmHTMRGso5uxqIKqptvUYVYIn9cUeo6Nm9rx96XTU/edit#heading=h.jbxd7bgt1gts)

n.b. "views" and "downloads" here apply to _any_ platform that might embed the object, not just purl pages, which is why the actual statistics collection has to happen in sul-embed.

- [x] display the number of times an object has been viewed
- [x] display the number of times an object has been downloaded
- [x] add a link to some explanatory info (in a modal?) about how metrics are collected. see [zenodo faq](https://help.zenodo.org/faq/#statistics)

### user stories
_As a depositor, I would like to be able to easily find out how many times my deposit to the SDR has been viewed and/or downloaded so that I can demonstrate the impact of my work to funders, promotion and tenure committees, potential employers, etc._

_As the Office of Scholarly Communication, I would like to be able to point to examples of SDR items that are extensively accessed or downloaded so that I can illustrate the value of open access deposits in the SDR._

_As a SUL staff member, I would like to easily find out how many times the SDR content I manage for my collections has been viewed and/or downloaded so that I can make better collection development and management decisions._

_As a self-deposit collection manager, I want to know which items in my collection are being viewed and downloaded the most so I can understand which content is the most useful to people._

_As a researcher who lands on a PURL page, I want to know how often this item has been viewed or downloaded because this information serves as a sort of first-level proxy for the value of the content._

_As a funder who is looking at published works of an applicant, I want to know how often the item has been viewed or downloaded so that I can include that in an assessment of its broader impact._

### design
![Screenshot 2023-10-18 at 15 41 06](https://github.com/sul-dlss/purl/assets/4924494/cb876dd1-c91b-4f0b-a85e-e909522f452a)

[See full design on invision](https://projects.invisionapp.com/share/FV13CCPQXMHA#/screens/474686424)

thatbudakguy:
One area we might still need some clarity on: do we care about showing "unique" views and downloads? In previous discussions, we talked about how it could be confusing for a user to see different numbers for "unique" and nonunique statistics, and wonder about the reasons that might be true (we probably won't know the answer).
- Is it even useful for people to deal with nonunique stats?
- If we decide to only show one category of stats, how does that change the suggested design?

thatbudakguy:
copying from slack:
[...]

PR #760, opened 2023-10-31
Display view counts on purl pages

thatbudakguy:
ref #721 

this adds a turbo frame for rendering metrics info after the main purl page has loaded.

Screenshot 2023-11-16 at 14 56 06


lwrubel:
Needs production deploy to unblock.

ndushay (approved):
LGTM;  nice and clean

PR #762, opened 2023-11-01
Get rid of print CSS.

justinlittman:
refs #757

Before:
Print CSS --> 283.71 KB
App CSS --> 282.74 KB
[bb051dp0564 - before.pdf](https://github.com/sul-dlss/purl/files/13227214/bb051dp0564.-.before.pdf)

After:
App CSS --> 284.86 KB
[bb051dp0564-after.pdf](https://github.com/sul-dlss/purl/files/13227218/bb051dp0564-after.pdf)


app/assets/stylesheets/print/styles.scss
```diff
@@ -1,91 +1,93 @@
...
-.panel-heading {
-  padding-left: 0;
-}
-#masthead {
-  border: none;
-}
-#su-content {
-  padding-bottom: 0!important;
-}
+  h3.index_title {
```
lwrubel, line 35: Maybe outside the scope of this PR, but where is this style even used?

lwrubel (approved):

lwrubel:
Based on the PDFs you shared, the new print styles are also a printing improvement by using up more of the page.

PR #763, opened 2023-11-01
Adjust heading order.

### LINKED ISSUE #755: Page should start with an H1 tag (and headings should be not out of order)
On the [PURL landing page](https://purl.stanford.edu/) (only page affected), **About Persistent URLs (PURLs)** should be an H1 instead of an H2. And since headings shouldn't be out of order or skip a level, that means that ### Featured Items should be an H2 instead of an H3.

Reference: https://www.w3.org/WAI/tutorials/page-structure/headings/

Screenshot 2023-10-31 at 1 22 53 PM

justinlittman:
closes #755

The easiest possible way to do this is change the html heading tag but add the bootstrap heading class, e.g., `<h2>About Persistent URLs (PURLs)</h2>` --> `<h1 class="h2">About Persistent URLs (PURLs)</h1>`. However, this means that some of the `<h1>` properties are picked up, changing the layout slightly.

Before:
image

After:
image

Fixing this would require some CSS shenanigans. Considering this is a landing page that no one actually lands on (I believe), I'd suggest that it isn't worth extra CSS complexity.

@dbranchini Is this acceptable?

### DIFF
--- a/app/views/purl/index.html.erb
+++ b/app/views/purl/index.html.erb
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-md-8">
-    <h2>About Persistent URLs (PURLs)</h2>
+    <h1 class="h2">About Persistent URLs (PURLs)</h1>
 
     <p>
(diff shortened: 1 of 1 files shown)


jcoyne (approved):

justinlittman:
@dbranchini approved.

PR #767, opened 2023-11-01
Remove popper js from importmap.rb

justinlittman:
To the best I can tell it isn't used.


jcoyne:
popper.js is a peer dependency of bootstrap.  Can you remove bootstrap too?

justinlittman:
If I'm not mistaken, all of those BS components are opt-in. And I don't see any indication that we've opted in to any of them.

jcoyne:
@justinlittman I don't see anything about opting in here: https://getbootstrap.com/docs/5.3/components/dropdowns/ All you have to do to opt in is to add the attribute `data-bs-toggle="dropdown"` to a DOM node.

justinlittman:
Dropdowns, popovers, and tooltips aren't used in purl. However, since it is a tiny library, I'm going to close.

Issue #774, opened 2023-11-02
Determine work needed to allow crawling of certain purl pages


amyehodge:
Determine what work would be needed to intentionally allow crawling of purl pages that have world downloadable content. 

At this time, we have not evaluated the implications of crawling other content enough to be comfortable allowing the crawling of anything else. When that analysis has been done and a policy decision made, we will revisit whether further content should be crawlable.

justinlittman:
To accomplish this, noindex meta tag should be added to excluded items (https://developers.google.com/search/docs/crawling-indexing/block-indexing) and they should not be included in the sitemap.

lwrubel:
@justinlittman: What will allow crawling of newly published, world-downloadable content? Will future content be included in the sitemap?

edsu:
One thing stakeholders might want to consider is whether we want our robots.txt to allow content to be accessible by GenerativeAI bots like [OpenAI](https://platform.openai.com/docs/gptbot) and [Google's Bard](https://blog.google/technology/ai/an-update-on-web-publisher-controls/). This is markedly different from providing content to be indexed, which drives traffic to PURL to view the published content, rather than using the content to answer questions directly, without attribution.

amyehodge:
Additional questions that we would like answered to see if we can get some content opened up for crawling:
- Can we exclude specific collections if we know some that we do not want to have crawled, even if they are world view/world download?
- Can we restrict crawling to only certain agents (see @edsu 's question above) to say, allow Google search crawling but nothing else? And then whitelist other agents can be added as we see fit.
- Can we set this up so that only content that comes in via self-deposit is crawled?

justinlittman:
> Can we exclude specific collections if we know some that we do not want to have crawled, even if they are world view/world download?
Yes

> Can we restrict crawling to only certain agents?
Yes

> Can we set this up so that only content that comes in via self-deposit is crawled?
Maybe. How would this content be identified? What about library content submitted via H2?

peetucket:
Just to note that directives to crawlers in robots.txt can be ignored by malicious actors.  While we expect major players like Google to follow our requests, we need to manage expectations with our stakeholders that other AI systems or search engine may still index or harvest content that is public even when we asked them not to.

amyehodge:
@justinlittman regarding

> Maybe. How would this content be identified? What about library content submitted via H2?

[...]

PR #776, opened 2023-11-03
Add noindex meta when not world download.

justinlittman:
This is part of opening up world download items to google indexing. It specifically prevents the opposite -- indexing items that are not world download by adding a noindex meta tag.

refs #774

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,3 +50,9 @@ The RuboCop style enforcement can be run without running the tests
 ## Deploying
 
 Deployment is handled automatically via Jenkins when a release is published to GitHub.
+
+## Search engine indexing
+
+The current policy is to only index items that have world view and download rights.
+
+To this end, only world view and download right items are included in the sitemap and all other items have a "noindex" meta tag.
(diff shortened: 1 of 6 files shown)


cbeer:
Is this the right implementation for #774? It looks like this would exclude lots of content (stanford-only, world-no-download, etc) from search engines, and it's not clear to me why we want to explicitly exclude those. If that's intentional, could the decision making (and shift from "exclude citation only" to "only world-download") be documented somewhere?

amyehodge:
@cbeer I have updated the original ticket. There has not been a shift in the decision, it was just poorly worded. Please let me know if the explanation there is still lacking.

edsu (approved):
Perhaps it's obvious but since this doesn't close a specific ticket I think the PR could use a description of what we are hoping to achieve by adding this metadata to the page?

justinlittman:
Comment updated @edsu

jcoyne:
I think this design goal should be added on the Readme too.  Something like: "PURL objects that have world accessable content files are offered for indexing by search engines"

justinlittman:
Good idea. Added @jcoyne .

PR #777, opened 2023-11-03
Create a sitemap.

justinlittman:
refs https://github.com/sul-dlss/purl/issues/774

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -1,5 +1,6 @@
 # Set the host name for URL creation
 SitemapGenerator::Sitemap.default_host = Settings.sitemap.default_host
+SitemapGenerator::Sitemap.sitemaps_path = 'public/system/sitemap/'
 
 SitemapGenerator::Sitemap.create(include_root: false) do
   # Put links creation logic here.
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -1,0 +1,33 @@
+# Set the host name for URL creation
+SitemapGenerator::Sitemap.default_host = Settings.sitemap.default_host
+
+SitemapGenerator::Sitemap.create(include_root: false) do
+  # Put links creation logic here.
+  #
+  # The root path '/' and sitemap index file are added automatically for you.
+  # Links are added to the Sitemap in the order they are specified.
+  #
+  # Usage: add(path, options={})
+  #        (default options are used if you don't specify)
+  #
+  # Defaults: :priority => 0.5, :changefreq => 'weekly',
+  #           :lastmod => Time.now, :host => default_host
+  #
+  # Examples:
+  #
+  # Add '/articles'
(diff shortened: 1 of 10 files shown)


config/sitemap.rb
```diff
@@ -0,0 +1,32 @@
...
+  # Add '/articles'
+  #
+  #   add articles_path, :priority => 0.7, :changefreq => 'daily'
+  #
+  # Add all articles:
+  #
+  #   Article.find_each do |article|
+  #     add article_path(article), :lastmod => article.updated_at
+  #   end
+  PurlResource.all.each do |purl|
```
cbeer, line 27: How long does it take to create a sitemap for -prod? My memory is iterating through `PurlResource` takes... days?

justinlittman: Testing that now. I went to query the PURL db, but, oh wait ... (just kidding).

justinlittman: In UAT it took 36 hours.

justinlittman:
If merging, please also undraft and merge https://github.com/sul-dlss/shared_configs/pull/2103

peetucket:
It's unclear if the combination of disallowing all crawling and a sitemap will work.  See https://webmasters.stackexchange.com/questions/139014/only-allow-crawling-of-pages-in-sitemap

But I don't know what the solution is... basically to explicitly and only allow specific URLs to be indexed.

peetucket:
> It's unclear if the combination of disallowing all crawling and a sitemap will work. See https://webmasters.stackexchange.com/questions/139014/only-allow-crawling-of-pages-in-sitemap
> 
> But I don't know what the solution is... basically to explicitly and only allow specific URLs to be indexed.

Oh, so is the solution that items that should *not* be indexed will have something added to each PURL, as noted in #774 and https://developers.google.com/search/docs/crawling-indexing/block-indexing  ?  And then we open up robots.txt to crawling ?  And if so, do we need some process to add that that meta tag to PURLs that should not be indexed?

justinlittman:
Not following @peetucket. There is a header added to each purl that should not be indexed (anything that is not world world). Every that should be indexed (anything that is world world) is included in the sitemap. Once we have the go-ahead, the do not index will be removed from `robots.txt`.

peetucket:
> Not following @peetucket. There is a header added to each purl that should not be indexed (anything that is not world world). Every that should be indexed (anything that is world world) is included in the sitemap. Once we have the go-ahead, the do not index will be removed from `robots.txt`.

I see - I missed that the no-index header is already in PURLs that need it, I though that still needed to be figured out (both how to identify these PURLs and then the code to add the appropriate header).  If both of those are already done, concur with your plan.

lwrubel (approved):

lwrubel:
Approved pending rebase.

PR #787, opened 2023-11-06
Avoid redundant calls to purl when fetching oEmbed.

justinlittman:
refs https://github.com/sul-dlss/purl/issues/757

### DIFF
--- a/app/javascript/controllers/oembed_controller.js
+++ b/app/javascript/controllers/oembed_controller.js
@@ -12,36 +12,12 @@ export default class OembedController extends Controller {
       return
     }
 
-    fetch(this.urlValue)
-      .then((response) => {
-        if (response.ok) return response.text()
-        throw new Error(`HTTP error, status = ${response.status}`)
-      })
-      .then((body) => {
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/oembed_controller.js
```diff
@@ -12,24 +12,37 @@ export default class OembedController extends Controller {
...
 
+    if(this.urlValue === document.URL) {
+      const oEmbedEndPoint = document.head.querySelector('link[rel="alternate"][type="application/json+oembed"]')?.getAttribute('href')
+      if (!oEmbedEndPoint) {
+        console.warn(`No oEmbed endpoint found in <head> at ${this.urlValue}`)
+        return
+      }
+      this.loadEndPoint(oEmbedEndPoint)
+    } else {
+      this.fetchAndLoadOEmbedEndpoint()
```
jcoyne, line 23: Why do we need this path?

justinlittman: I don't think we do, but I had thought you were suggesting that we needed to support arbitrary oEmbed. Let me take it out.

jcoyne (approved):

PR #805, opened 2023-11-08
Add missing require to schedule.rb.

justinlittman:


jcoyne (approved):

config/schedule.rb
```diff
@@ -18,6 +18,7 @@
 # end
 
 # Learn more: http://github.com/javan/whenever
+require_relative 'environment'
 require 'config'
```
jcoyne, line 22: Do we still need this?

mjgiarlo: I don't believe we do.

justinlittman: Agree, based on GB. Removed.

Issue #761, opened 2023-11-01
Refactor: Move IIIF NavPlace generation to iiif-prezi/osullivan


justinlittman:
As a follow on to #752, add NavPlace support to https://github.com/iiif-prezi/osullivan and remove generation from Iiif3PresentationManifest:NavPlace.

See https://github.com/sul-dlss/purl/pull/752#discussion_r1377942226

PR #821, opened 2023-11-16
Use navPlace from iiif-presentation gem

peetucket:
Part of https://github.com/sul-dlss/purl/issues/761 : moves navPlace generation from PURL (added here: https://github.com/sul-dlss/purl/pull/752) to the iiif-presentation gem, as well as removes knowledge of public-xml from the class that generates it (and instead leaves that knowledge in PURL).

~~HOLD since this requires a new release of the iiif-presentation gem first, currently in PR: https://github.com/iiif-prezi/osullivan/pull/86~~

This existing spec verifies that it continues to produce the expected output: https://github.com/sul-dlss/purl/blob/main/spec/features/iiif3_manifest_spec.rb#L385-L416

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,6 @@ gem 'cancancan' # authorization
 gem 'config' # simple rails environment specific config
 gem "cssbundling-rails", "~> 1.1"
 gem 'faraday' # HTTP client
-gem "geo_coord", require: "geo/coord"
 gem 'honeybadger' # exception reporting
 gem 'htmlentities'
 gem "importmap-rails" # Use JavaScript with ESM import maps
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -31,7 +30,7 @@ gem 'whenever', require: false # cron jobs
 
 # DLSS and its community
 gem 'dor-rights-auth', '~> 1.6'
-gem 'iiif-presentation', '~> 1.2'
+gem 'iiif-presentation', '~> 1.2', git: 'https://github.com/peetucket/osullivan', branch: 'add-nav-place-support'
 gem 'mods_display', '~> 1.4'
```
peetucket, line 34: for tests to work, until new release of gem is available

PR #826, opened 2023-11-20
Various pins to use rack 2.

### LINKED ISSUE #820: purl-feedback should submit message instead of producing error
When I try to submit feedback on a purl page, I get the following when I click the "Send" button:

Screenshot 2023-11-16 at 10 03 33 AM

This happens when I have filled in all three fields on the form and am:
- not logged in, so I have to do the reCAPTCHA  
- logged in, so I have to check the box for reCAPTCHA instead of picking tiles out of a photo
- logged in and skip checking the box for reCAPTCHA

justinlittman:
Phusion does not yet support rack 3.

closes #820


justinlittman:
After fix:
image

PR #835, opened 2023-11-28
Shorten the distance between metrics headings and counts

thatbudakguy:
requested by @amyehodge via slack. since we're only showing the one metric right now (and eventually only two), having the big padding in between the heading and count looks a little weird.

before:

Screenshot 2023-11-28 at 09 02 34

after:

Screenshot 2023-11-28 at 09 02 30

### DIFF
--- a/app/assets/stylesheets/modules/record-metadata-section.scss
+++ b/app/assets/stylesheets/modules/record-metadata-section.scss
@@ -46,9 +46,10 @@
 }
 
 .metrics dl {
-  max-width: 40ch;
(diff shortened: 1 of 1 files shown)


justinlittman:
Should this using the forthcoming table layout? It seems odd to be trying to layout a dl to look like a row.

justinlittman:
I think you'll find the need for this is obviated with the new design.

amyehodge:
@justinlittman Agreed.

Issue #842, opened 2023-12-01
Consider refactor of section html


justinlittman:
![image](https://github.com/sul-dlss/purl/assets/588335/cb1c1105-87bf-45d7-a6d0-970e47c3f00b)

Issue #799, opened 2023-11-07
Update design for PURL pages


amyehodge:
See designs below.

Item PURL example: https://projects.invisionapp.com/share/FV13CCPQXMHA#/screens/474703516
Item PURL with everything we could think of on it for clarity on the order of things when they do exist on the page: https://projects.invisionapp.com/share/FV13CCPQXMHA#/screens/474842888
Collection PURL example: https://projects.invisionapp.com/share/FV13CCPQXMHA#/screens/474718726

justinlittman:
What do you mean by "Item PURL with everything we could think of on it for clarity on the order of things when they do exist on the page"? The linked image for an existing purl page, right?

justinlittman:
Can you please provide sample markup for the new layout styles that demonstrate the proper accessibility markup?

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVc5QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c433a7bd6223a15aae67a411b0a62d0edfdd448b/image.png)

and 

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWk5QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--98319283586083deb065e8fe129606ed81870995/image.png)

justinlittman:
Can you provide designs on how the columns should re-flow on smaller screens?

amyehodge:
> What do you mean by "Item PURL with everything we could think of on it for clarity on the order of things when they do exist on the page"? The linked image for an existing purl page, right?

Different kinds of items include different metadata. Instead of trying to create a mockup of every possible combination of metadata, which is impossible, @astridu tried to pull together as many of the different metadata fields as possible to show them all on one page. No PURL looks like the example "frankenPURL," but it's intended to serve as a reference for where sections go in relation to each other when they are present.

amyehodge:
@astridu Can you answer @justinlittman 's other questions above? Thanks.

justinlittman:
> > What do you mean by "Item PURL with everything we could think of on it for clarity on the order of things when they do exist on the page"? The linked image for an existing purl page, right?
> 
[...]

PR #843, opened 2023-12-01
Adds new section and table layout components.

justinlittman:
refs #799

Note that this uses the existing html for sections. For a proposed refactor of this see #842.


jcoyne (approved):
This looks good. Do you want to update the templates to use this component too?

justinlittman:
@jcoyne I thought about adding that to the PR, but it was going to make a mess of the redesign PR.

PR #849, opened 2023-12-04
Temporary placeholder before the embed loads.

jcoyne:
This prevents the content from jumping around


justinlittman (approved):
Approve pending review by @astridu.

PR #877, opened 2023-12-07
Refactor link style to be additive rather than subtractive

jcoyne:
This fixes the double underline in the global footer when hovering.

### DIFF
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -29,7 +29,3 @@
 .record-metadata {
   margin-bottom: 2em;
 }
-
-.altmetric-embed a {
-  border-bottom: none;
-}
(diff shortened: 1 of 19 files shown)


jcoyne:
@justinlittman because there are some links on the page where we don't want it. Namely the ones in the global footer where we currently see a double underline.  We also don't want it on links that are images, or in the header.

justinlittman:
Could it just be limited to links in the relevant parts of the page? My concern is that a future dev will add a new link but not think of adding the necessary class.

jcoyne:
@justinlittman at this point it's harder to handle the exceptions rather than to remember to add the one style we want to use. So long as we are using border-bottom for a link style, this gets very challenging.  I think if we went to text-decoration:underline, I could get behind a link style default.

justinlittman:
I defer to your judgment. Thanks for the explanation.

justinlittman (approved):

PR #893, opened 2023-12-12
Allow crawling in robots.txt

justinlittman:


ndushay (commented):
will this keep "private" purls private due to sitemap?  If not, we can't do this.

lwrubel:
Blocking indexing on everything else is handled by: https://github.com/sul-dlss/purl/blob/1cdb714d869f74169e07adf044ace4425918a2db/app/views/purl/show.html.erb#L17

public/robots.txt
```diff
@@ -1,7 +1,7 @@
 # See http://www.robotstxt.org/wc/norobots.html for documentation on how to use the robots.txt file
 #
 # To ban all spiders from the entire site uncomment the next two lines:
-User-Agent: *
-Disallow: /
```
lwrubel, line 5: I think we want to Allow crawling, rather than Disallow?

justinlittman: This is commenting out the Disallow, which should translate into an implicit Allow.

lwrubel: Got it.

lwrubel (approved):

PR #895, opened 2023-12-12
Rescue errors in schema.org generation

lwrubel:
Rescue errors in thumbnailUrl generation or schema.org markup as a whole.

### DIFF
--- a/lib/metadata/schema_dot_org.rb
+++ b/lib/metadata/schema_dot_org.rb
@@ -20,6 +20,9 @@ module Metadata
         "description": description }
         .merge(format_specific_fields)
         .compact
+    rescue StandardError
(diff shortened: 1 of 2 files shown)


justinlittman (commented):
Can you rescue something more specific than StandardError?

lwrubel:
I made the error rescued in the `thumbnail` method more specific: `URI::InvalidURIError` since that's the likely error raised with an unexpected jp2 filename (e.g. with special characters) or some other issue creating the URL. 

I'm not sure what other specific error to rescue at the level of `#call`. A problem with schema.org metadata generation should not keep the PURL page from rendering, so that's what I'm trying to avoid. Are there other errors you'd suggest I consider (or something else I might try)?

lwrubel:
Added.

justinlittman (approved):

PR #896, opened 2023-12-12
Stop using sass variables

jcoyne:
We can use css properties now.


edsu (approved):
I guess at some point we don't need the .scss file extension?

justinlittman (approved):

PR #909, opened 2023-12-13
Display tracked download metrics

### LINKED ISSUE #901: Show download metrics for downloadable objects
The SDR metrics API already returns counts of download events; https://github.com/sul-dlss/sul-embed/issues/1935 will result in us actually tracking them once deployed.

We want to add a new row to the table of displayed metrics for all object types that can be downloaded:

- geo
- pdf
- files
- 3d
- media

Web archives can't be downloaded (at least not through the embed viewer), so we shouldn't show the download count on those purl pages.

We can't track downloads for IIIF images right now because it'd likely require changes to a Mirador plugin.

thatbudakguy:
Closes #901


app/views/purl/_metrics.html.erb
```diff
@@ -2,12 +2,22 @@
   <% label_id = 'section-metrics' %>
   <%= render SectionComponent.new(label: 'Usage metrics', label_id: label_id) do %>
     <% if document.metrics.present? %>
-      <%= render TableComponent.new(label_id: label_id) do %>
+      <%= render TableComponent.new(label_id: label_id, with_body: false) do %>
+      <tbody>
         <tr>
           <th>Views</th>
-          <td><%= number_with_delimiter document.metrics['unique_views'] %></td>
+          <td class="view-count"><%= number_with_delimiter document.metrics['unique_views'] %></td>
```
justinlittman, line 9: Suggest using an id or data attribute rather than a class, since this is being used solely for testing, not formatting.

justinlittman (approved):
Approve pending consideration of suggestion.

app/views/purl/_metrics.html.erb
thatbudakguy: Makes sense to me – changed it to an id

PR #912, opened 2023-12-13
Revert "Remove sass"

jcoyne:
This reverts commit 5fd22919260636c110cf1e841103855ecafebd0a.

Some of the older < 120 versions of chrome and < 17.2 of safari don't have full support yet.
See https://caniuse.com/css-nesting


jcoyne:
I think all the features we use *are* supported.

PR #914, opened 2023-12-14
Update metrics date message

### LINKED ISSUE #902: Update note about metrics effective date
We currently show a notice on PURL pages indicating when metrics collection began:

Screenshot 2023-12-12 at 13 59 33

This notice is redundant for any content deposited after the end of the workcycle, since all of its metrics will be captured. We also need to update the notice to take account of when downloads started being tracked.

- [ ] Hide the notice if content was deposited after the end of the workcycle (12/15/23)
- [ ] Change the notice to a more general time ("December 2023") or update it to mention specifically that download tracking began later than view tracking did

lwrubel:
Resolves #902 by having one date for both views and downloads, on all PURLs. 

@amyehodge here's the updated message:

Screenshot 2023-12-14 at 10 54 36 AM

### DIFF
--- a/app/views/purl/_metrics.html.erb
+++ b/app/views/purl/_metrics.html.erb
@@ -22,7 +22,7 @@
       <p>No usage metrics available.</p>
     <% end %>
     <p>
-      Data collected since 11/20/23.
+      Data collected since 12/23.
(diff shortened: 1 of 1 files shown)


lwrubel:
@amyehodge here's the updated message.

justinlittman (approved):

PR #935, opened 2024-01-05
Update integration spec document cache from production

jcoyne:
So that our integration tests continue to track what we have in production.


spec/features/integration_spec.rb
```diff
@@ -139,16 +139,7 @@
       visit '/tx027jv4938'
       expect(page).to have_content 'IMSA 24 Hours of Daytona'
       expect(page).to have_metadata_section 'Access conditions'
-      expect(page).to have_metadata_section 'Description'
```
jcoyne, line 142: We removed all of our revs specific descriptions and moved all of this back to the Revs institute. See https://purl.stanford.edu/tx027jv4938

thatbudakguy (approved):

PR #954, opened 2024-02-05
Don't advertize IIIF manifests for collections

### LINKED ISSUE #952: [PURL/prod] NoMethodError: undefined method `at_xpath' for nil:NilClass
## Backtrace

    line 19 of [PROJECT_ROOT]/app/models/content_metadata.rb: book_data
    line 15 of [PROJECT_ROOT]/app/models/content_metadata.rb: reading_order
    line 5 of [PROJECT_ROOT]/app/models/iiif3_presentation_manifest.rb: reading_order

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49629/faults/103754969)

jcoyne:
Ensure they are always 404 rather than a 500 error

Fixes #952

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -12,6 +12,9 @@ class IiifController < ApplicationController
     # See ContentMetadata::GroupedResource#primary
     return head :not_found if @purl.type == 'geo'
 
+    # Avoid trying to create a manifest for collections, because they have no files, and thus do not validate.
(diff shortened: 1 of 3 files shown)


cbeer:
Instead of guarding against collections, shouldn't we just guard against bad data (and presumably catch other types of objects that are missing content metadata too)?

https://github.com/sul-dlss/purl/pull/955

jcoyne:
@cbeer I believe this does both.

PR #956, opened 2024-02-06
Produce stub IIIF collection manifests for SDR collections

cbeer:
Includes #955

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -29,7 +29,7 @@ class Iiif3PresentationManifest < IiifPresentationManifest
       manifest_data['@context'] = IIIF::V3::Presentation::CONTEXT + ['http://iiif.io/api/extension/navplace/context.json']
     end
 
-    manifest = IIIF::V3::Presentation::Manifest.new manifest_data
+    manifest = iiif_manifest_class.new(manifest_data)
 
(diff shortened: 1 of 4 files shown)


app/models/iiif3_presentation_manifest.rb
```diff
@@ -29,7 +29,7 @@ def body
       manifest_data['@context'] = IIIF::V3::Presentation::CONTEXT + ['http://iiif.io/api/extension/navplace/context.json']
     end
 
-    manifest = IIIF::V3::Presentation::Manifest.new manifest_data
+    manifest = iiif_manifest_class.new(manifest_data)
 
     # Set viewingHint to paged if this is a book
     manifest.viewingHint = 'paged' if type == 'book'
```
jcoyne, line 35: Don't we want some way to short circuit this method if the resource is a collection, so we don't end up with a viewingDirection, etc. https://github.com/sul-dlss/purl/blob/83628a26521368918dae575f273331c3913b720d/app/models/iiif3_presentation_manifest.rb#L46C14-L46C30

jcoyne (changes requested):

app/models/iiif3_presentation_manifest.rb
cbeer: 🤷‍♂️ 

> A Collection may have the viewingDirection property.

jcoyne (approved):

PR #957, opened 2024-02-07
Use text-underline-position rather than a bottom-border for link styles

jcoyne:
Using bottom-border was a hack we devised years ago before `text-underline-position: under` was an option.  This keeps the underline from intersecting the descenders.


corylown (approved):
This change seems okay to me. It does push the underline down below the text quite a bit, but seems to match the always underlined links:

Prod:
Screenshot 2024-02-22 at 9 56 38 AM

This PR:
Screenshot 2024-02-22 at 9 56 57 AM


Always underlined links on prod/this PR:
Screenshot 2024-02-22 at 9 55 11 AM

Issue #959, opened 2024-02-16
Add more collections to the allow list for web crawlers


amyehodge:
The 150 H2 collection druids that should be added to the allowed list to be made available to web crawlers is:

cp667df5882
vk630sh8038
sz556bc6971
pn596gn7571
km092my4610
mz061vb5322
bs335mc1127
hn807kb3349
tn384hm9336
fb707zy6875
pn826mx7934
xs573nh8717
yp108tn6462
bd442kg4610
xk076fr1089
cb368sb3215
xf370bb6498
zp761vy5985
xn235kn0506
jk216wd8778
kv041ng6595
mp757fq0674
tp622mj4093
qk912vj7408
js114xn4901
pp784wp5100
cj269gn0736
hg577nd6300
df918cs3825
pm409fz5494
qd259nz6640
tz959sb6952
yk425tq4374
gy156rk1135
pn808wc6253
nc035fh7626
dc746gw6209
vr801mm4929
sm380jb1849
mz198fp9366
wp446bg1851
vj354pj0087
xg006bv4120
dc052nc3046
zh458bn3243
nj669yc9949
jm681xz5805
wg919by6182
kk382jq8063
db700wz4340
pz310tn8238
fh113fq5916
pn159pj7895
dn774gf4795
bs190xh6401
rp102sx8543
kq629sd5182
tx773dg8618
ff984ns9763
kx581zg6755
zm381kw7359
rr255gf6892
qc836mt3783
ns899tx9783
gs860rq5377
zp689kv8227
rf106nf0893
pf670qd3191
mm346cv9110
vx401mn2884
cm018rf5314
zj027sq8811
pv178fc4399
nr325kt8685
wt021dc3029
dn308bq6123
qy070zf4368
tk177rs9408
ph261dt3685
dh544bj2035
wh620hw5859
ky703mk0439
ym197kw4639
dz588pg3041
jq136sm3689
cw267yk6233
cz636rc5886
ky865kd2888
yd955cs5721
dm470nm9547
qx425hy3186
qr032th6424
gb482gq5743
fd702hd2985
cm344zy0053
xx313vx3974
yf927nn8171
rh525wj3800
md903dt5665
tn741mc5549
vj318ph3055
th767br8285
wc569dd1362
sn758bh0099
gw182vg0580
cs911pc6181
dr680zx7328
rm471fw5693
sv179hy0727
kh840qv1626
cd881bx9575
vs366nb6695
rv822xv5878
xf974wp1364
[...]

PR #962, opened 2024-02-21
add 150 h2 crawlable collections to settings.yml

ndushay:
Part of #959

@lwrubel is this what you meant?


config/settings.yml
```diff
@@ -99,6 +99,157 @@ crawlable:
     - kp887kn2294
     # BIOHOPK185 Student videos
     - km092my4610
+    # H2 (self deposit) collections in the allowed list to be made available to web crawlers
```
lwrubel, line 102: All of `crawlable.collections` are to be made available to web crawlers, so I'm not sure you need that part of the comment.

lwrubel (approved):

config/settings.yml
ndushay: I revised the comment, thanks!  Will merge when green.

PR #973, opened 2024-04-16
Build the sitemap from the release tags

### LINKED ISSUE #971: Use release tags to determine what is included in sitemap
* Change PurlResource.crawlable? to use "PURL sitemap" release tag from public XML to determine what is added to sitemap.
* Remove the existing setting.
* Run test sitemap generation to make sure it is working properly.

jcoyne:
Fixes #971

### DIFF
--- a/app/models/purl_resource.rb
+++ b/app/models/purl_resource.rb
@@ -54,6 +54,16 @@ class PurlResource
     @public_xml ||= PublicXml.new(public_xml_document)
   end
 
+  def public_json
+    @public_json ||= purl_fetcher_conn.get("/purl/#{id}").body
+  end
+
+  def purl_fetcher_conn
+    Faraday.new(url: Settings.purl_fetcher.url) do |builder|
+      builder.response :json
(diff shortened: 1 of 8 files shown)


app/models/purl_resource.rb
```diff
@@ -70,9 +80,9 @@ def containing_purl_collections
 
   # Can be crawled / indexed by a crawler, e.g. Googlebot
   def crawlable?
```
justinlittman, line 82: Can you get rid of this method?

justinlittman (changes requested):

app/models/purl_resource.rb
jcoyne: It's called in `app/views/purl/show.html.erb`

justinlittman: Ah, OK.

justinlittman (approved):

justinlittman:
This has been tested so merging.

PR #999, opened 2024-05-20
Refactor 'View in SearchWorks' & 'View in EarthWorks' links

thatbudakguy:
Previously, we prevented the rendering of EarthWorks view links
for all content that didn't have the geo content type. We now have
a lot of scanned maps in EarthWorks that deserve to be linked to
from Purl.

The logic for deciding how and when to render these links was also
convoluted, so this PRs simplifies it somewhat and expands the
tests.

### DIFF
--- a/app/views/purl/_find_it.html.erb
+++ b/app/views/purl/_find_it.html.erb
@@ -1,22 +1,18 @@
 <% releases = [] %>
 
 <% Settings.releases.each do |dest| %>
-  <% next if dest[:key] == "Searchworks" && document.catalog_key %>
-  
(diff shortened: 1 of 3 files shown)


thatbudakguy:
You can test this with `ct435ky3217`, which currently doesn't show a "view in EarthWorks" link despite being released to EarthWorks.

thatbudakguy:
This also appears to fix a bug where `xy271vj8397` currently shows a "view in EarthWorks" link when it shouldn't.

jcoyne (approved):

PR #1002, opened 2024-05-21
Determine release status by calling the API

### LINKED ISSUE #1000: Decouple from release tags in XML
Because metadata will be versionable, and release tags are not part of a version, we need to look elsewhere for release tags.

https://github.com/sul-dlss/purl/blob/82055e8a1fbb95893c87998084c8f3cbb5dc32be/app/models/public_xml.rb#L33-L37

jcoyne:
release is no longer in the versioned metadata

fixes #1000

### DIFF
--- a/app/models/preview_resource.rb
+++ b/app/models/preview_resource.rb
@@ -18,7 +18,6 @@ class PreviewResource
     []
   end
 
-  def released_to?(_key); end
   def copyright?; end
   def use_and_reproduction?; end
   def catalog_key; end
--- a/app/models/public_xml.rb
+++ b/app/models/public_xml.rb
@@ -30,12 +30,6 @@ class PublicXml
     @folio_instance_hrid ||= document.root.at_xpath('identityMetadata/otherId[@name="folio_instance_hrid"]')&.text.presence
   end
 
-  def released_to?(key)
-    release = document.root.at_xpath("releaseData/release[@to='#{key}']")&.text
-
-    release == 'true'
-  end
-
   def thumb
     document.root.at_xpath('thumb')&.text
   end
--- a/app/models/purl_resource.rb
+++ b/app/models/purl_resource.rb
@@ -80,10 +80,15 @@ class PurlResource
 
   # Can be crawled / indexed by a crawler, e.g. Googlebot
   def crawlable?
-    public_json.fetch('true_targets').include?('PURL sitemap')
-  rescue StandardError => e
-    Honeybadger.notify(e)
-    false # ensure that purl-fetcher being down doesn't prevent the page from drawing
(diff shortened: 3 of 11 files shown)


cbeer:
I guess I'm puzzled why we want PURL to make API calls instead of using data from disk. That seems like a major architectural departure (although I see that ship sailed with `crawlable?`) and at the very least suggests purl-fetcher should become a load-balanced, high-availability application. I'm not sure why we can't keep reading data from disk though.. that pattern seems like it's served us well.

jcoyne:
@cbeer it's due to the decoupling of versionable data and non-versionable data.  It would have been an option to write the non-versionable data to a disk, but the disk is also connected via a network, so I don't see why a new shared drive with a bunch of uncontrolled data is better than an API.

cbeer:
>  It would have been an option to write the non-versionable data to a disk

That seems like a good idea and keeps with the existing architecture. I'm pretty sure our experience has shown that file systems are generally a little more stable than full application stacks (and presumably have some built-in redundancy 🤷‍♂️) and easier to debug. It also seems like it'd be way easier to develop on PURL if our fixture data is static and we don't have to set up a full purl-fetcher stack too.

jcoyne:
The HA idea is a good one and has already been discussed. I've created the ticket to push forward on that: https://github.com/sul-dlss/operations-tasks/issues/3760

jcoyne:
>  I'm pretty sure our experience has shown that file systems are generally a little more stable than full application stacks (and presumably have some built-in redundancy 🤷‍♂️) and easier to debug. It also seems like it'd be way easier to develop on PURL if our fixture data is static and we don't have to set up a full purl-fetcher stack too.

I agree with some of this, but we've also found the data on disk to be unvalidated while data that gets stored in a database and processed by some kind of schema validator (i.e. in an API) tends to have a more predictable structure, which leads to more robust applications.

cbeer:
I thought Cocina data was validated before being written? Shouldn't that enforce data validation? Doing it after it hits PURL seems too late.

I don't think we're talking about a lot of data, and one of the core architectural principles behind PURL has been to limit its run-time dependencies as much as possible. I don't see why we'd throw that away for something this trivial, at least.

jcoyne:
> I thought Cocina data was validated before being written? Shouldn't that enforce data validation? Doing it after it hits PURL seems too late.

Cocina is validated, but it is the "versioned metadata". The release status is "non-versionable metadata"

> I don't think we're talking about a lot of data, and one of the core architectural principles behind PURL has been to limit its run-time dependencies as much as possible. I don't see why we'd throw that away for something this trivial, at least.

We have had this in production for a couple of months now I believe.  Let's discuss it at standup tomorrow if we want to re-think how we'll move forward.

PR #1007, opened 2024-05-23
Read files from OCFL root when feature flag is enabled

### LINKED ISSUE #997: Update purl to look for metadata in the OCFL root
~~Blocked by https://github.com/sul-dlss/ocfl-rb/issues/17~~

Using a feature flag, enable PURL to read `public` xml from the OCFL root (`/stacks/ocfl/<druid tree>/v1/content/') rather than the `/purl` root.  Put this behind a feature flag.

mjgiarlo:
Fixes #997

This commit makes it possible for Purl to read files from an OCFL-structured storage root, and puts that logic behind a new feature flag. To accomplish this, it encapsulates knowledge of OCFL vs. legacy document cache structures within a new `OcflPathFinder` class.


app/models/purl_resource.rb
```diff
@@ -269,7 +273,13 @@ def last_modified_header_value
 
   concerning :ActiveModelness do
     def attributes
-      { druid: id, druid_tree: }
+      {
+        druid: id,
+        druid_tree:,
+        root_path: self.class.storage_root_path,
+        ocfl_public_path: OcflPathFinder.path(druid_tree:, filename: 'public'),
+        ocfl_cocina_path: OcflPathFinder.path(druid_tree:, filename: 'cocina.json')
```
mjgiarlo, line 281: Instead of a more radical change, use the current strategy for figuring out paths to `public` and `cocina.json` files.

mjgiarlo, line 281: These are different paths because they may come from different OCFL versions.

spec/fixtures/ocfl/bb/000/br/0025/0=ocfl_object_1.1
mjgiarlo, line 1: To help the tests pass when the new flag is on, migrate all the current fixtures into an OCFL structure (leaving the current fixtures where they are).

spec/model/purl_resource_spec.rb
```diff
@@ -38,7 +38,7 @@
 
     it 'validates that the object is "ready"' do
       allow_any_instance_of(described_class).to receive(:public_xml?).and_return(false)
-      expect { described_class.find('bc000df0000') }.to raise_error PurlResource::ObjectNotReady
+      expect { described_class.find('bc421tk1152') }.to raise_error PurlResource::ObjectNotReady
```
mjgiarlo, line 41: Using a real druid here because otherwise this fails when the flag is on.

jcoyne (approved):

PR #1012, opened 2024-05-29
Simpler meta format

jcoyne:


justinlittman (approved):
Can you please add documentation about this to README?

PR #1028, opened 2024-06-12
Prefer directly calling the component to using a helper.

jcoyne:
This reduces indirection


cbeer:
If we're saying this is the preferred form, should we also update https://github.com/sul-dlss/purl/blob/main/app/views/purl/_mods_bibliographic.html.erb and remove the helper?

jcoyne:
@cbeer do you mean this PR? https://github.com/sul-dlss/purl/pull/1029

thatbudakguy (approved):

PR #1030, opened 2024-06-13
Extract subject component

jcoyne:


app/components/subject_component.html.erb
```diff
@@ -0,0 +1,15 @@
+<% label_id = "section-subjects" %>
```
mjgiarlo, line 1: If this value doesn't vary, could it be returned by a method in the Ruby half of the component instead of being a template-local variable?

app/components/subject_component.html.erb
```diff
@@ -0,0 +1,15 @@
...
+<%= render SectionComponent.new(label: 'Subjects', label_id:) do %>
+  <%= render TableComponent.new(label_id:) do %>
+    <tbody>
+      <% document.mods.subject.each do |field| %>
+        <%= render ModsDisplay::RowFieldComponent.new(field:, value_transformer: ->(subject_line) { safe_join(link_mods_subjects(subject_line), ' > ') }) %>
```
mjgiarlo, line 6: Do you want these using the row field component from the gem or the (now local) component?

app/components/subject_component.html.erb
jcoyne: It could be, but this is just moving the existing code from [app/views/purl/_mods_subjects.html.erb](https://github.com/sul-dlss/purl/pull/1030/files#diff-3e775b6ccf9e8669ae7a03ec24e888cadb838ea27c217577953230b461a16893). I think that might cause more indirection for little benefit.

mjgiarlo (approved):

PR #1032, opened 2024-06-17
Set CONTROLMASTER_HOST for Jenkins

jcoyne:


cbeer:
I think https://github.com/sul-dlss/dlss-capistrano/pull/50 will be a better fix (and/or, isn't it really weird to have dlss-ci-prod connect to itself in order to deploy out to other stuff?)

PR #1054, opened 2024-06-26
Split resource and version

jcoyne:


app/views/purl/_collection_items.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if document.collection? && document.released_to_searchworks? %>
```
justinlittman, line 1: Why did you get rid of this?

justinlittman (approved):
Approve pending consideration of question.

app/views/purl/_collection_items.html.erb
jcoyne: It's moved outside the partial because `released_to_searchworks?` is on the PurlResource rather than the PurlVersion.

PR #1055, opened 2024-06-27
Refactor PurlResource to extract ResourceRetriever

jcoyne:
The PurlResource was doing too many different responsibilities, this removes one.


app/models/purl_resource.rb
```diff
@@ -66,13 +65,10 @@ def metrics
...
   end
 
-  # For building links in _find_it.html.erb
-  def attributes
-    {
-      druid:,
-      druid_tree:,
-      root_path: self.class.storage_root_path
-    }
+  delegate :meta_json_body, :updated_at, to: :resource_retriever
```
justinlittman, line 68: While we're refactoring can we call meta_json_body something clearer? I have no idea what it might be based on the variable name.

justinlittman (approved):
Approve pending consideration of comment.

app/models/purl_resource.rb
jcoyne: Suggestions? It's the body of the meta.json

justinlittman: Oh, that's right. Maybe just a comment reminding the dev what meta_json contains?

PR #1066, opened 2024-07-09
Fix the meta_json request

jcoyne:
PurlResource#meta_json is called by the PurlsController, so it needs to be public

### DIFF
--- a/spec/request/purl_spec.rb
+++ b/spec/request/purl_spec.rb
@@ -25,6 +25,22 @@ RSpec.describe 'PURL API', type: :request do
         expect(response.body).to include "The page you were looking for doesn't exist."
       end
     end
+
+    context 'for json' do
+      it 'returns the Cocina json' do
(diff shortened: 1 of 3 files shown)


spec/request/purl_spec.rb
```diff
@@ -25,6 +25,22 @@
...
+        expect(response.parsed_body).to include('cocinaVersion', 'type', 'structural')
+      end
+    end
+
+    context 'for meta_json' do
+      it 'returns the meta.json' do
+        get '/bb157hs6068.meta_json'
+        expect(response).to be_successful
+        expect(response.parsed_body).to include('cocinaVersion', 'type', 'structural')
+      end
```
peetucket, line 42: copypasta on the test expectation 

and a rubocop complaint

peetucket (commented):
looks good other than test/rubocop issues

peetucket (approved):

PR #1068, opened 2024-07-11
Add versions section behind a query parameter feature flag

jcoyne:
Feature flag is:
http://localhost:3000/bb112zx3193?version_feature=1


app/components/versions_component.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= render SectionComponent.new(label: 'Versions', label_id:) do %>
+  <%= render TableComponent.new(label_id:) do %>
+    <% versions.each do |version| %>
+      <tbody>
+        <tr>
+          <th>Version 1</th>
```
aaron-collier, line 6: Should this and line 8 be updated to reflect the version instead of status `Version 1`? Or are these static on purpose?

jcoyne: Yes, these are static on purpose as it gives us the ability to react to and critique the layout even though we don't yet have source data for the list of versions.

aaron-collier (approved):

PR #1078, opened 2024-07-24
Add route to retrieve a purl with a version ID

### LINKED ISSUE #1039: Update display to use versioned metadata
With versioned metadata available, the purl display will need to be updated to use the metadata for the version provided, or the latest version if not provided.

Add an optional version parameter to the purl URL that will be used for direct linking into a version and citations with version information.

mjgiarlo:
Fixes #1039

The only UI change is this banner when viewing any version other than the head version:

Screenshot from 2024-07-24 14-06-04

This does not implement the UI as specified in the Figma. Instead, it uses the existing code for handling alerts, for consistency reasons.


app/controllers/purl_controller.rb
```diff
@@ -12,9 +12,11 @@ def index; end
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/MethodLength
   def show
-    @version = @purl.version(:head)
+    @version = @purl.version(version_param)
     return unless stale?(last_modified: @version.updated_at.utc, etag: @version.cache_key + "/#{@version.updated_at.utc}")
 
+    flash[:alert] = 'A newer version of this item is available' unless @version.head?
```
cbeer, line 18:
```suggestion
    flash.now[:alert] = 'A newer version of this item is available' unless @version.head?
```

I think?

app/controllers/purl_controller.rb
```diff
@@ -12,9 +12,11 @@ def index; end
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/MethodLength
   def show
-    @version = @purl.version(:head)
+    @version = @purl.version(version_param)
```
cbeer, line 15: What should we do if the version doesn't exist? (Presumably not a 500 error: https://app.honeybadger.io/projects/49629/faults/109769206)

mjgiarlo: I'll whip up a test for this and handle the error.

app/controllers/purl_controller.rb
mjgiarlo: This is what Rubocop wanted, but I didn't see any usage of `flash.now` in purl so I was hesitant to patch that in. Safe enough to try and test it.

cbeer: Huh. I thought `flash[]` was for redirects, and `flash.now` was for rendering. Maybe that's not cool any more 🤷‍♂️

mjgiarlo: @cbeer the request spec is happy with the change at least.

justinlittman (approved):

PR #1084, opened 2024-07-26
Retreive meta.json from stacks for versioned objects

### LINKED ISSUE #1083: When versioned layout, get meta.json from stacks
Currently, meta.json is written to purl file system for all objects. In addition, it is also written to stacks file system for versioned objects (<object directory>/versions/meta.json). In the future, meta.json will not be written to the purl file system.

When reading meta.json, if the object is versioned, get it from the stacks file system; otherwise, get it from the purl file system.

Here is existing code for determining if versioned: https://github.com/sul-dlss/purl/blob/088fb87ac17d0af6aa1f235670e068ac241b1aae/app/services/resource_retriever.rb#L69-L71

aaron-collier:
Fixes #1083 

When an object is versioned, retrieve meta.json from the stacks path. Corresponding shared_configs PRs:

https://github.com/sul-dlss/shared_configs/pull/2228 (stage)
https://github.com/sul-dlss/shared_configs/pull/2229 (uat)
https://github.com/sul-dlss/shared_configs/pull/2230 (prod)

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -24,6 +24,7 @@ purl_resource:
   versioned:
     public_xml: "https://purl.stanford.edu/%{druid}/version/%{version_id}.xml"
     cocina: "https://purl.stanford.edu/%{druid}/version/%{version_id}.json"
+    meta: "https://purl.stanford.edu/%{druid}.meta_json" # versioned meta is not supported
 
 embed:
   url: "https://purl.stanford.edu/%{druid}"
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -24,7 +24,7 @@ purl_resource:
   versioned:
     public_xml: "https://purl.stanford.edu/%{druid}/version/%{version_id}.xml"
     cocina: "https://purl.stanford.edu/%{druid}/version/%{version_id}.json"
-    meta: "https://purl.stanford.edu/%{druid}.meta_json" # versioned meta is not supported
+    meta: "https://purl.stanford.edu/%{druid}/meta.json" # versioned meta is not supported
 
 embed:
   url: "https://purl.stanford.edu/%{druid}"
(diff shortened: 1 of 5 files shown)


config/settings/test.yml
```diff
@@ -6,6 +6,7 @@ purl_resource:
   versioned:
     public_xml: "%{root_path}/%{druid_tree}/%{druid}/versions/public.%{version_id}.xml"
     cocina: "%{root_path}/%{druid_tree}/%{druid}/versions/cocina.%{version_id}.json"
+    meta: "%{root_path}/%{druid_tree}/meta.json"
```
mjgiarlo, line 9: Can you also add a line like this to the README? There's a section that documents how to do local dev.

mjgiarlo, line 9: I'm not sure how much it matters here, but the shared_configs changes expect `meta.json` in a `%{druid_tree}/%{druid}/versions/meta.json`. Should we reflect that structure in the test config so it's more similar, like we do with public XML and cocina?

aaron-collier: Good catch, I meant to update that, thanks.

config/settings.yml
```diff
@@ -24,6 +24,7 @@ purl_resource:
   versioned:
     public_xml: "https://purl.stanford.edu/%{druid}/version/%{version_id}.xml"
     cocina: "https://purl.stanford.edu/%{druid}/version/%{version_id}.json"
+    meta: "https://purl.stanford.edu/%{druid}.meta_json" # versioned meta is not supported
```
mjgiarlo, line 27: Should the `.` after the druid be a `/`?

aaron-collier: Ah, yes, but apparently not used... Blah. Fixing.

mjgiarlo (approved):

PR #1094, opened 2024-08-02
Return minimal cocina with status withdrawn for withdrawn versions

### LINKED ISSUE #1091: Do not display full public cocina/xml for withdrawn versions
When a version is withdrawn, we show a tombstone page: https://sul-purl-stage.stanford.edu/tm662fj8161/v1

We should not show the whole cocina or public xml, which is still being served: https://sul-purl-stage.stanford.edu/tm662fj8161/v1.json

aaron-collier:
Fixes #1091

### DIFF
--- a/app/controllers/purl_controller.rb
+++ b/app/controllers/purl_controller.rb
@@ -38,7 +38,9 @@ class PurlController < ApplicationController
       end
 
       format.json do
-        if @version.cocina?
+        if @version.withdrawn?
+          render json: withdrawn_cocina
+        elsif @version.cocina?
           render json: @version.cocina_body
         else
           head :not_found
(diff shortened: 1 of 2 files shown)


app/controllers/purl_controller.rb
```diff
@@ -108,4 +110,23 @@ def maybe_add_flash_message!
...
+    cocina_json = JSON.parse(@version.cocina_body)
+
+    {
+      cocinaVersion: cocina_json['cocinaVersion'],
+      type: cocina_json['type'],
+      externalIdentifier: cocina_json['externalIdentifier'],
+      label: cocina_json['label'],
+      version: cocina_json['version'],
+      status: 'withdrawn',
+      administrative: {},
```
justinlittman, line 126: I'd recommend getting rid of cocinaVersion, administrative, identification, description, and structural.

jcoyne: Won't this break stacks when it tries to parse it?

jcoyne: https://github.com/sul-dlss/stacks/blob/main/app/models/cocina.rb#L81

justinlittman: Shouldn't stacks handle something that is withdrawn explicitly (look for the status param) instead of implicitly?

Happy to discuss, as maybe I'm misunderstanding something.

jcoyne: If we just provide an empty structural, then there is no need to change stacks.

justinlittman (approved):

PR #1137, opened 2024-08-15
Don't include version in sul-embed iframe URL when purl is base url

### LINKED ISSUE #1135: Embed code should not use versioned URL when purl URL is not versioned
When there's a "base" purl without a version in the URL, the embed code should also not use a version in the URL. It looks like embed is currently putting the latest version number in the URL.

Screenshot 2024-08-14 at 4 46 24 PM

Example: https://sul-purl-stage.stanford.edu/fz115cn9991

embed URL when viewing that page: https://embed-stage.stanford.edu/iframe?url=https://sul-purl-stage.stanford.edu/fz115cn9991/version/2

lwrubel:
Resolves #1135.

Updates embed iframe URL and embed code to use base URL when head version. 

Andrew tested and approved.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -34,10 +34,10 @@ module ApplicationHelper
   def embeddable_url(druid, version_id = nil)
     if Settings.embed.url
       (Settings.embed.url % { druid: }).tap do |embed_url|
-        return "#{embed_url}/version/#{version_id}" if version_id.present?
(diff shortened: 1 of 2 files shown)


lwrubel:
@andrewjbtw would you like to test this?

andrewjbtw:
Yes - is it already in stage?

lwrubel:
Requires a tweak so that if the URL is a head version with a version in the URL, still include that in the embed URL. The presence or lack of a version in the URL should be the same as what's in the embed URL.

justinlittman (approved):

PR #1144, opened 2024-08-19
Extract a flash view_component

jcoyne:


app/components/flash_message_component.rb
```diff
@@ -0,0 +1,30 @@
...
+    @flash = flash
+    super
+  end
+
+  def render?
+    !@flash.empty?
+  end
+
+  def each_type
+    [:success, :notice, :error, :alert].each do |level|
```
corylown, line 16: Did you intend to use `TYPES` here instead?

corylown (changes requested):

app/components/flash_message_component.rb
jcoyne: Yes! Thanks Cory. I've fixed that.

corylown (approved):

PR #1156, opened 2024-09-11
Add URL for right-click actions

### LINKED ISSUE #1152: "copy url" link in versions widget is incorrect in some browsers and contexts
There's a "Copy URL" link in the versions widget in the Purl sidebar:

Screenshot 2024-08-28 at 3 33 06 PM

Left-clicking on this link should copy the URL for that specific version to the user's clipboard. This seems to be functioning ok in Firefox, Chrome, and Safari.

However, right-clicking on the link is leading to the wrong URL in some browsers.

Example from: https://sul-purl-stage.stanford.edu/fv934js5171

The "copy url" link on this page should go to:
`https://sul-purl-stage.stanford.edu/fv934js5171/version/1`

When left-clicking, that's the URL I get for my clipboard in Chrome, Safari, and Firefox on a Mac.

But when I right-click I get:

- Firefox: `https://sul-purl-stage.stanford.edu/fv934js5171#`
- Chrome: `https://sul-purl-stage.stanford.edu/fv934js5171/version/1#`
- Safari: `https://sul-purl-stage.stanford.edu/fv934js5171/version/1#`

Firefox appears to be the outlier in not linking correctly to the versioned URL. I'm also not sure why there's a `#` in all of those URLs from right-clicking.

lwrubel:
Resolves #1152 so that right-clicking and copying also gets the versioned URL in all browsers. 

@andrewjbtw tested to confirm working as expected.

### DIFF
--- a/app/components/version_actions_component.html.erb
+++ b/app/components/version_actions_component.html.erb
@@ -4,7 +4,7 @@
   <% else %>
     <%= link_to 'View', version %>
   <% end %>
-  | <a href="#" data-action="click->clipboard#copy">Copy URL</a>
+  | <%= link_to 'Copy URL', version_purl_url(id: version.druid, version: version.version_id), data: { action: "clipboard#copy"} %>
 <% else %>
   <%= state.capitalize %>
 <% end %>


app/components/version_actions_component.html.erb
```diff
@@ -4,7 +4,7 @@
   <% else %>
     <%= link_to 'View', version %>
   <% end %>
-  | <a href="#" data-action="click->clipboard#copy">Copy URL</a>
+  | <a href="<%= version_purl_url(id: version.druid, version: version.version_id) %>" data-action="click->clipboard#copy">Copy URL</a>
```
justinlittman, line 7: Can you change this to a `link_to`?

lwrubel: Done

app/components/version_actions_component.html.erb
```diff
@@ -4,7 +4,7 @@
   <% else %>
     <%= link_to 'View', version %>
   <% end %>
-  | <a href="#" data-action="click->clipboard#copy">Copy URL</a>
+  | <%= link_to 'Copy URL', version_purl_url(id: version.druid, version: version.version_id), data: { action: "click->clipboard#copy"} %>
```
justinlittman, line 7: FYI -- "click" isn't necessary.

justinlittman (approved):

app/components/version_actions_component.html.erb
lwrubel: Thanks--not sure I realized I could drop that in converting to `link_to`. I'll tweak it and then wrap it up.

lwrubel: Oh, cool. Thanks for explaining.

PR #1168, opened 2024-11-21
Update to the latest component library

jcoyne:


thatbudakguy (approved):

dnoneill:
You might want to wait on merging. I think there are still questions if we are going to update all the header/footers at the same time or piecemeal.

jcoyne:
@dnoneill Then we're stuck with this bug? Until when?  Screenshot 2024-12-03 at 8 33 07 AM

dnoneill:
I don't know. I would check with Vivian.

cbeer:
I don't see why fixing the bug and doing a total branding change have to be connected. Can't we just add some CSS to fix it in that app?
https://github.com/sul-dlss/purl/pull/1172

jcoyne:
"total branding change" is just a change of the icon.  This was asked for by the DLUX team: https://github.com/sul-dlss/component-library/issues/55

dnoneill:
Sounds like this can be merged.

PR #1172, opened 2024-12-03
Turn off text-decoration for the Stanford logo to match upstream.

cbeer:


app/assets/stylesheets/application.css
```diff
@@ -125,3 +125,7 @@
 .section-header {
   color: var(--stanford-cardinal);
 }
+
+.su-logo {
```
jcoyne, line 129: This is already fixed in https://github.com/sul-dlss/component-library/commit/07217a97d4988e607eadf8892717858391fac466 along with some other style discrepancies.

PR #1182, opened 2025-01-21
update mobile menu to open below site title

dnoneill:
Screenshot 2025-01-21 at 3 17 39 PM
Screenshot 2025-01-21 at 3 17 32 PM

### DIFF
--- a/app/views/shared/_brand_navbar.html.erb
+++ b/app/views/shared/_brand_navbar.html.erb
@@ -5,7 +5,7 @@
       <span class="navbar-toggler-icon"></span>
     </button>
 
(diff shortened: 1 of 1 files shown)


app/views/shared/_brand_navbar.html.erb
```diff
@@ -31,5 +31,14 @@
         <div class="h1 my-3"><a href="/">Stanford Digital Repository</a></div>
       </div>
     </div>
+    <div class="navbar collapse navbar-collapse" aria-label="browse" id="user-util-collapse">
```
jcoyne, line 34: Do we need aria-label here?

PR #1193, opened 2025-02-10
Create IIIF manifests for geo objects

jcoyne:


dnoneill (approved):
Screenshot 2025-02-10 at 4 01 48 PM

PR #1196, opened 2025-02-12
hide_metadata parameter is obsolete

jcoyne:
See https://github.com/sul-dlss/SearchWorks/pull/1084


dnoneill:
@jcoyne can you update the commit message to be hide_metadata instead of this? This isn't really descriptive

jcoyne:
@dnoneill sure, done.

PR #1215, opened 2025-02-24
Build iiif metadata directly from cocina

### LINKED ISSUE #1214: Display "Abstract" as a header in the sul-embed info panel
To support display of self-deposited research, I would like "Abstract" to appear as a header in the sul-embed info panel when an item has an abstract. Currently, there's no label on the abstract in that panel.

> Image

This may require modifying the mapping from MODS to DC. Currently, the fields are published like this, with only MODS tagging an abstract as an abstract:

DC: <dc:description>This is the abstract field</dc:description>
MODS: <abstract>This is the abstract field</abstract>

jcoyne:
This fixes the problem where we can't express 'abstract' in DC elements 1.1, and makes it unnecessary for us to continue to publish (broken) DC in the future

Fixes #1214


app/models/iiif3_presentation_manifest.rb
```diff
@@ -36,10 +36,12 @@ def body
 
     # Set viewingHint to paged if this is a book
     manifest.viewingHint = 'paged' if type == 'book'
-
-    metadata_writer = Iiif3MetadataWriter.new(dc_nodes: public_xml_document.xpath('//oai_dc:dc/*', 'oai_dc' => OAI_DC_SCHEMA),
-                                              published_dates: public_xml_document.xpath('/publicObject/@published').map(&:text),
-                                              url: controller.purl_url(druid))
+    collection = containing_purl_collections&.first
+    collection_title = collection.second.title if collection
```
dnoneill, line 40: leave comment about the shape of the object

dnoneill (approved):

PR #1228, opened 2025-03-05
fix manifest for pdfs

### LINKED ISSUE #1226: [PURL/prod] NoMethodError: undefined method 'id' for nil
## Backtrace

    line 128 of [PROJECT_ROOT]/app/models/iiif3_presentation_manifest.rb: Iiif3PresentationManifest#canvas_for_resource
    line 96 of [PROJECT_ROOT]/app/models/iiif3_presentation_manifest.rb: block in Iiif3PresentationManifest#build_canvases
    line 95 of [PROJECT_ROOT]/app/models/iiif3_presentation_manifest.rb: Array#each

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49629/faults/117940744)

dnoneill:
closes #1226

### DIFF
--- a/app/models/content_metadata.rb
+++ b/app/models/content_metadata.rb
@@ -60,8 +60,9 @@ class ContentMetadata
       return files.first if files.length == 1
 
       return media_file if media_file.present?
+      return image_file if image_file.present?
 
-      image_file
+      pdf_file
     end
 
     def supplementing_resources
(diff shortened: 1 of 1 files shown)


app/models/iiif3_presentation_manifest.rb
```diff
@@ -182,8 +182,9 @@ def supplementing_resources_annotation_page(resource_group)
   end
 
   def renderings_for_resource_group(resource_group)
+    ### renderings should not have probe services
```
jcoyne, line 185: Can you say why they should not? Is it a spec violation, or just because we don't need this info?

dnoneill: I think it is a spec violation. https://iiif.io/api/presentation/3.0/#rendering doesn't say anything about being able to have a service and it kind of doesn't make sense with what it is.

dnoneill: Thanks I missed that. I will go ahead and remove this change.

PR #1230, opened 2025-03-06
don't build items for file types

### LINKED ISSUE #1229: Don't create items list in IIIF manifest for file type objects.
File type objects can have thousands of files on them, which makes the iiif manifest very large and take a long time to generate (each file has all the services for access, etc).   At present we only use the metadata from the manifest, so we can leave the items list empty.  Please add comments in the code indicating why we made this choice.

dnoneill:
closes #1229

### DIFF
--- a/spec/requests/iiif3_manifest_spec.rb
+++ b/spec/requests/iiif3_manifest_spec.rb
@@ -264,7 +264,7 @@ RSpec.describe 'IIIF v3 manifests' do
     end
   end
 
-  context 'with a PDF object' do
+  context 'with a file object' do
     let(:druid) { 'bb132pr2055' }
 
     it 'generates a correct manifest' do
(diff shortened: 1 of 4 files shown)


spec/requests/iiif3_manifest_spec.rb
```diff
@@ -298,22 +293,10 @@
...
-      expect(canvas['height']).not_to be_present
-      expect(canvas['width']).not_to be_present
 
-      pdf = canvas['items'].first['items'].first
-      body = pdf.fetch('body')
-      expect(body['id']).to eq 'https://stacks.stanford.edu/file/bb253gh8060/SC0193_Agenda_6381_2010-10-07_001.pdf'
-      expect(body['format']).to eq 'application/pdf'
-      expect(body['type']).to eq 'Text'
-
-      probe_service = body.fetch('service').first
```
jcoyne, line 312: These were the only tests that the probe service is added to a manifest. Can you keep these?

jcoyne (changes requested):

spec/requests/iiif3_manifest_spec.rb
dnoneill: I can add it to something else but this druid is a file not a pdf so I don't think it is appropriate with this druid.

jcoyne: That's right.  You might need to add a document or media type.

dnoneill: Done

PR #1234, opened 2025-03-12
Account for structured values in notes

### LINKED ISSUE #1233: Viewer info panel: table of contents is blank
I've found items where the descriptive metadata contains a table of contents but the contents are not in the metadata in the viewer: https://purl.stanford.edu/dc682mb8493

Image

I see contents in the Dublin Core, so I think it was previously included in manifests:
```
    <dc:description displayLabel="Contents">Nelson Mandela (Bambata's Children of Nathal) -- Intro (Aswad's Brinsley Forde) -- The gun (Benjamin Zaphaniah) -- We're coming long way/Not last (Young Power) -- Intro (Aswad's Brinsley Forde) -- Solidarity (Rocka's Delight) -- Want Fi Goh Rave (Linton Kwesi Johnson and Dennis Bovell Dub Band) -- Freely (Bob Andy and Friends -- What are we doing? (Twinkle Brothers)</dc:description>
```

corylown:
Fixes #1233

### DIFF
--- a/app/models/iiif3_metadata_writer.rb
+++ b/app/models/iiif3_metadata_writer.rb
@@ -93,9 +93,7 @@ class Iiif3MetadataWriter
   end
 
   def resource_types
-    filtered_form('resource type').flat_map do
-      it['structuredValue'].presence ? it['structuredValue'].map { it['value'] } : it['value']
-    end.uniq(&:downcase)
+    filtered_form('resource type').flat_map { structured_values(it) }.uniq(&:downcase)
   end
 
   def filtered_form(type)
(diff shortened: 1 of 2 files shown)


dnoneill:
@corylown I think this is going to be happening for other fields. Can we make a shared function. I know right now title is using the same functionality. Can we make something shared?

corylown:
It didn't look like there was enough uniformity across how we're handling each field to extract something generic.

dnoneill:
@corylown I think https://github.com/sul-dlss/purl/blob/6fe08d4d9b43586ff2522b80fed80fb27ee4ae31/app/models/iiif3_metadata_writer.rb#L97 might be able to have a shared function. I am not totally sure but I think it could. The one different is the list for non structured values but I think it would still work.

corylown:
@dnoneill the difference is that for notes we're casting `it[''value]` to an array, but in the method you showed we're using flat_map so we don't need to do that.

dnoneill:
@corylown this looks good. Can you add a test for resource_type? It should already be in the cocina_descriptive so you just need a test.

corylown:
@dnoneill I believe `resource type` is already tested via the tests for "Type" which is where that `resource_types` method is used. 

- https://github.com/sul-dlss/purl/blob/6fe08d4d9b43586ff2522b80fed80fb27ee4ae31/app/models/iiif3_metadata_writer.rb#L76
-

dnoneill:
Okay I see what is happening. I got confused because the purl.find is overwriting the default. Thanks.

dnoneill (approved):

Issue #84, opened 2016-09-09
Generate IIIF collection URI's for image collections


snydman:
We should generate collection level IIIF URI's for our defined image collections, following the spec:

http://iiif.io/api/presentation/2.1/#collection

A top level collection manifest of collections would be good too.

aeschylus:
Here is an example from Jeff Witt's project: http://scta.info/iiif/collection/scta

Some other quasi-collections: http://iiif.archivelab.org/iiif/collection.json

aeschylus:
Context: https://groups.google.com/forum/#!topic/iiif-discuss/r1ANaZG-BQo

cbeer:
I'm not sure we currently publish enough information to PURL for it to a) identify an image collection, and b) build a collection manifest that lists the members of that collection.

cbeer:
But the collection doesn't know its items?

thatbudakguy:
> purl-fetcher should have the data to query for members of a collection.

would it be a problem to have purl do this at render time when generating the manifest?

jcoyne:
Given that Purl caches the manifests, I think it should be okay to do.  Out of curiosity, why are we tackling this now?

thatbudakguy:
> why are we tackling this now?

We aren't, I'm mostly just thinking through some things (and maybe creating speculative tickets) based on conversations coming out of a conference we're hosting right now.

andrewjbtw:
Collections can have items of any type. In practice, there are collections that contain only images, but the only way to know that would be to query the type on each item. Items of other types could be added at any time. There are very good reasons from an intellectual organization standpoint to keep SDR collections type-agnostic.

thatbudakguy:
> the only way to know that would be to query the type on each item

it seems like the way to address that would be modifying purl-fetcher's `/collections/[:druid]/purls` endpoint to return content type in addition to object type, e.g. for https://purl-fetcher.stanford.edu/collections/ct961sj2730/purls:

```diff
{
  "purls": [
    {
      "druid": "druid:hg144mf6252",
      "object_type": "item",
+     "content_type": "map",
      "title": "Africa",
      "catkey": "a11400706",
      "true_targets": [
        "Searchworks",
        "Earthworks",
        "SearchWorksPreview",
        "ContentSearch"
      ],
      "collections": [
        "druid:ct961sj2730"
      ],
      "updated_at": "2025-02-27T21:44:41Z",
      "published_at": "2025-02-27T21:44:41Z",
[...]

PR #1238, opened 2025-03-25
Fetch collection membership and populate IIIF collection manifests

thatbudakguy:
Ref #84

This uses the new capabilities of purl-fetcher's collections
endpoint to populate and filter a list of collection members,
and then add their manifest URLs to the items list for the
collection's manifest.

Member items are filtered based on their content type and release
status, so that only publicly released items with IIIF-displayable
content types are added. The actual rights of the object are
delegated to the viewer if the manifest is loaded.

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -5,7 +5,7 @@ class Iiif3PresentationManifest < IiifPresentationManifest
   delegate :reading_order, to: :content_metadata
   attr_reader :purl_base_uri
 
-  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity
   def body
     manifest_data = {
       'id' => manifest_url,
@@ -53,6 +53,7 @@ class Iiif3PresentationManifest < IiifPresentationManifest
     manifest.thumbnail = [thumbnail_resource] if thumbnail_resource?
 
     build_canvases(manifest)
+    build_collection_items(manifest) if collection?
 
     manifest
   end
@@ -93,6 +94,17 @@ class Iiif3PresentationManifest < IiifPresentationManifest
       content_metadata.grouped_resources.each do |resource_group|
         manifest.items << canvas_for_resource(resource_group)
       end
+    end
+  end
+
+  def build_collection_items(manifest)
+    # for each member of the collection, add its manifest to the collection's items
+    iiif_collection_members.each do |member|
+      manifest.items << {
+        'id' => manifest_url(id: member['druid'].delete_prefix('druid:')),
+        'label' => { en: [member['title']] },
+        'type' => 'Manifest'
+      }
     end
   end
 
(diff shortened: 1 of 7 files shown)


thatbudakguy:
# Collection manifests in Mirador
Screenshot 2025-03-25 at 15 38 52

Screenshot 2025-03-25 at 15 38 33

thatbudakguy:
# Collection manifests in Universal Viewer 4

You get a sidebar similar to Mirador's that lists all the collection items, except it's on the left and the selected manifest info is on the right.

It appears there is built-in sorting by date and volume (?) but we aren't populating these in the collection manifest, so not sure how it's supposed to work.

Screenshot 2025-04-01 at 13 25 52

thatbudakguy:
# Collection manifests in Clover

You immediately view the first manifest. There might be collection navigation in the top-left dropdown, but it didn't do anything when I clicked it in testing...

Screenshot 2025-04-01 at 13 27 51

app/models/iiif3_presentation_manifest.rb
```diff
@@ -441,4 +454,24 @@ def iiif_manifest_class
...
+      .select { |member| iiif_content_types.include? member['content_type'] }
+  end
+
+  private
+
+  # Include collection members released to any of these platforms
+  # TODO: create a release tag just for this purpose?
+  def iiif_release_targets
+    %w[Searchworks EarthWorks]
+  end
```
thatbudakguy, line 471: First-pass implementation of "is this public enough to be listed in a collection manifest"?

app/models/iiif3_presentation_manifest.rb
```diff
@@ -441,4 +454,24 @@ def iiif_manifest_class
...
+
+  # Filter out collection members that can't be rendered in a IIIF viewer
+  def iiif_content_types
+    %w[image book map]
+  end
```
thatbudakguy, line 476: I think Mirador and UV actually can render video and audio, so maybe we could expand this further?

dnoneill: They do. https://iiif.io/api/cookbook/recipe/0003-mvm-video/ https://iiif.io/api/cookbook/recipe/0002-mvm-audio/

app/services/collection_members_retriever.rb
```diff
@@ -0,0 +1,38 @@
...
+  #
+  # Alternatively, use ActiveRecord hooks in purl-fetcher’s database to
+  # update the collection’s timestamp when its members are modified
+  #
+  # For now, no caching until we know we need it
+  def collection_members
+    @collection_members ||= benchmark "Fetching #{druid} collection_members at #{collection_members_path}" do
+      purl_fetcher_client.collection_members(druid).to_a
+    end
+  end
```
thatbudakguy, line 20: The manifest caching we do in the controller seems to complicate this; in testing I had to append bogus params like `?resource_id=1` in order to force it to expire and render the new manifest that contained items

dnoneill:
@thatbudakguy I think you don't have a label or I am not sure what field they are looking at in clover, but it isn't set so if you click on the dropdown it should switch.
429210664-1ed32fb8-62bd-44b8-8129-a98b99bdec15

thatbudakguy:
🤷🏻 I couldn't get clicking that to do anything, so I gave up

PR #1239, opened 2025-03-26
Make IIIF v3 presentation provider into an array

thatbudakguy:
The spec (https://iiif.io/api/presentation/3.0/#provider) says
that this value must be an array of objects, but we currently
provide just a single object.

This causes the Clover viewer, in particular, to fail to render
manifests and complain that it can't .map() on the provider value.


jcoyne:
Will sul-embed be able to cope with this change?

jcoyne:
It looks like it can deal with it.

thatbudakguy:
Thanks for testing!

Issue #1242, opened 2025-04-02
Version 1 date display should reflect last system version date for v1 in Argo


amyehodge:
@andrewjbtw and I have noticed that purls seem to be displaying the last date the purl page was published for the Version 1 date in the versions section of the page. This should be displaying the most recent date for that system version in Argo. @andrewjbtw said "I think once you create a new user version, the date freezes for the old version, but until then it seems to float with the purl publish date rather than the last accessioned date." (Please let me know if that's not clear.)

Some examples, all of which have Jan 10, 2025 for the v1 date except for the one noted below:
https://purl.stanford.edu/bb333tj7983
https://purl.stanford.edu/bd485tn7072
https://purl.stanford.edu/bg091gk8548
https://purl.stanford.edu/bh082zh0439 (this one has Oct 18, 2024, but that's still way too recent)
https://purl.stanford.edu/bh364jv1083
https://purl.stanford.edu/bj100kq7976
https://purl.stanford.edu/bk962py0458
https://purl.stanford.edu/bq499mh5981
https://purl.stanford.edu/bq914sc3730
https://purl.stanford.edu/br087nj9017
https://purl.stanford.edu/by735ng7322
https://purl.stanford.edu/cg222yg9208
https://purl.stanford.edu/ch961jr6366
https://purl.stanford.edu/ck726bm9949

To illustrate, for the first purl above, I see this in the xml
`<publicObject id="druid:bb333tj7983" published="2025-01-10T03:20:58Z" publishVersion="cocina-models/0.98.1">`

this on the page

<img width="494" alt="Image" src="https://github.com/user-attachments/assets/f29ef443-cf53-450b-97fc-e8283920da7e" />

And this in Argo for the last system version (https://argo.stanford.edu/view/druid:bb333tj7983)

<img width="413" alt="Image" src="https://github.com/user-attachments/assets/29de9c5f-723e-4643-a242-ff96bb850c48" />

Here's an example of something with multiple versions that doesn't seem to have this problem: https://purl.stanford.edu/sq409qv3664

peetucket:
I think purl-fetcher is now producing the public xml from which the PURL page is pulling the publish date.  It looks like the date being shown on the purl page may the 'published' date at the root of the XML, which is set to the current date as of the time the public xml is generated.  see https://github.com/sul-dlss/purl-fetcher/blob/main/app/services/publish/public_xml_service.rb#L19

[...]

PR #1243, opened 2025-04-03
do not show version information unless there are user versions

peetucket:
See #1242 and https://stanfordlib.slack.com/archives/C09M7P91R/p1743613931774519

Basically, the `updated at` date we show for PURLs that have never had any user versions can be inaccurate, since it shows the last time the PURL was published, which can be more recent than the time the object was actually accessioned.

It may be possible to get this date in these cases from SDR through purl-fetcher and on to Purl, but it could be complicated (I'm not sure, but you can see some of my investigations in the ticket comments).  

Andrew suggested simply not showing the date in these cases since it was never shown before, and I think it probably doesn't even make sense to show the version panel either, since there isn't anything useful there.  My read of the code is that without user versions for the object, it will only ever show one version in that panel anyway, and the date could be wrong.  

See https://github.com/sul-dlss/purl/blob/main/app/models/purl_resource.rb#L21-L31 and https://github.com/sul-dlss/purl/blob/main/app/models/purl_resource.rb#L73-L93 (i.e. no version manifest?  no useful version info to display)

Verified on stage (by deploying this branch):
- example with user versions that still shows the version panel: https://sul-purl-stage.stanford.edu/bc670fy5418
- example without user versions that hides the version panel: https://sul-purl-stage.stanford.edu/bb011nz8100

ToDo:
- [x] test on stage
- [x] add tests

**example of before:**
Screenshot 2025-04-03 at 12 24 35 PM

**after:**
Screenshot 2025-04-03 at 12 25 00 PM

### DIFF
--- a/app/components/versions_component.rb
+++ b/app/components/versions_component.rb
@@ -10,11 +10,16 @@ class VersionsComponent < ViewComponent::Base
 
   attr_reader :purl, :label_id
 
+  # we only show version information if the object has user versions
+  def render?
+    purl.version_manifest_body.present?
+  end
+
   def embeddable_url
     @embeddable_url ||= helpers.embeddable_url(purl.druid)
   end
 
   def versions
-    @purl.versions.sort_by(&:version_id).reverse
+    purl.versions.sort_by(&:version_id).reverse
   end
 end
--- a/spec/features/display_purl_page_spec.rb
+++ b/spec/features/display_purl_page_spec.rb
@@ -100,15 +100,17 @@ RSpec.describe 'Displaying the PURL page' do
     end
   end
 
-  context 'revs object' do
+  context 'revs object (no version manifest)' do
     let(:druid) { 'tx027jv4938' }
 
-    it 'displays the page' do
+    it 'displays the page without version information' do
       visit "/#{druid}"
       expect(page).to have_content 'IMSA 24 Hours of Daytona'
       expect(page).to have_metadata_section 'Access conditions'
       expect(page).to have_metadata_section 'Bibliographic information'
       expect(page).to have_content 'Revs ID 2012-015GHEW-BW-1984-b4_1.4_0003'
+      expect(page).to have_no_content 'You are viewing this version'
+      expect(page).to have_no_content 'Each version has a distinct URL, but you can use this PURL to access the latest version.'
     end
   end
 
@@ -235,9 +237,17 @@ RSpec.describe 'Displaying the PURL page' do
     end
   end
 
-  context 'with a version' do
+  context 'with a version manifest' do
+    let(:druid) { 'wp335yr5649' }
+
+    it 'shows the version information panel' do
+      visit "/#{druid}"
+      expect(page).to have_content 'You are viewing this version'
+      expect(page).to have_content 'Each version has a distinct URL, but you can use this PURL to access the latest version.'
+    end
+
     it 'draws the page' do
-      visit '/wp335yr5649/version/3'
+      visit "/#{druid}/version/3"
       link = page.find('link[rel="alternate"][title="oEmbed Profile"][type="application/json+oembed"]', visible: false)
       expect(link['href']).to eq 'https://embed.stanford.edu/embed.json?url=https%3A%2F%2Fpurl.stanford.edu%2Fwp335yr5649%2Fversion%2F3'
     end


amyehodge:
I agree that if it's complex to sort out the accurate date and there is only ever one version in these situations, then it's better not to show the versions panel at all. Thanks for all the digging on this one.

app/components/versions_component.rb
```diff
@@ -10,11 +10,16 @@ def initialize(purl:, version:)
...
+    purl.version_manifest_body.present?
+  end
+
   def embeddable_url
     @embeddable_url ||= helpers.embeddable_url(purl.druid)
   end
 
   def versions
-    @purl.versions.sort_by(&:version_id).reverse
+    purl.versions.sort_by(&:version_id).reverse
```
peetucket, line 23: purl has an attribute reader

jcoyne:
I think we'd still want a link to version 1 even if there is no version manifest.  There may be one in the future, and we would want the user to be able to link to v1 if v2 shows up at a later time.

peetucket:
> I think we'd still want a link to version 1 even if there is no version manifest. There may be one in the future, and we would want the user to be able to link to v1 if v2 shows up at a later time.

If a user creates a version that is distinct enough to trigger display in this panel, won't the object then get a version manifest, and it will then display as expected?  That's what i understand from the code anyway, but this is my first time looking at this, would be good to verify.

peetucket:
In other words, once a version manifest becomes available, the purl page will revert to it's current behavior, which is to show the panel with all of the links.  All this does is suppress the display when there is nothing useful to show (i.e. no version manifest, and thus we'd always only show a single version with a possibly wrong date).

jcoyne:
> If a user creates a version that is distinct enough to trigger display in this panel, won't the object then get a version manifest, and it will then display as expected? 


Yes, you are correct about this, however, a person may want to have a link to v1 state even before they have triggered a version action, as they may not have control over, or insight into when v1 or v2 is released.  They just want it "as it is now".

peetucket:
> > If a user creates a version that is distinct enough to trigger display in this panel, won't the object then get a version manifest, and it will then display as expected?
> 
> Yes, you are correct about this, however, a person may want to have a link to v1 state even before they have triggered a version action, as they may not have control over, or insight into when v1 or v2 is released. They just want it "as it is now".

Oh I see.  While it would be useful to get back to that first "pre-user version" version, I'm actually not sure how this would work.  

I did a brief test using H2 on stage, and it doesn't seem like you can get back to your previous version even if you had that URL.  Here is what I did:

1. Found an older H2 object of mine before all of the versioning work.  It did not have any user versions/version manifest.  So in the current view, it would show as having a single version.  In this PR it would show nothing.
2. In H2, I versioned the object and asked it to make a new version.  I made some changes and hit deposit.
3. In the PURL page, I saw my update, but it still only showed one version with today's date (from the newly created version manifest).  So the display on current PURL wouldn't actually change except for the date shown, and you can't actually get back to the original version, even if you had that link from before.  In the code in this PR, it would start showing the version panel with the current version and date.

So I guess even if we did could just suppress the date instead of the whole panel, it's not really useful since I don't think we can get back to that very first version anyway?

jcoyne:
@peetucket I don't think there is such a thing as a pre-user version.  We consider v1 as the same as "you published it for the first time". e.g. https://sul-purl-stage.stanford.edu/bb011nz8100/version/1 is the same as https://sul-purl-stage.stanford.edu/bb011nz8100

peetucket:
Look at this H2 created object: https://argo-stage.stanford.edu/view/druid:jy904qg1183  

It was originally created by the H2 integration test in January 2024.  

Earlier today, I went into H2 and created a new user version of it (just changed some metadata by adding V2 to the title) and then deposited it.  Even though Argo shows both versions (the "original" version the integration test created via H2, and the new one I just adjusted today), the PURL page only shows a single version -- the latest one I did today: https://sul-purl-stage.stanford.edu/jy904qg1183 

You can't get back to the "previous" one done by the integration test in January 2024.  The only version URL that works is https://sul-purl-stage.stanford.edu/jy904qg1183/version/1 

I think this will be true of any object that had it's last version published before we started tracking user versions and publishing that version manifest?  So even if we display that version panel for these types of objects, copying that "current version URL" is not really even really accurate.  If a new user version of that older object is published, it will become the new "version 1".   At least as far as I can tell.  Perhaps this wasn't the intention, but it's what seems to be happening, at least in the test I did.

jcoyne:
@peetucket but then imagine drastically changing it for v2.  At least we'd have a way to get back to v1.   In the work cycle we said that we didn't want to support different "with version" and "without version" behavior.  I believe the intention is still to convert all non-versioned objects to objects with one user version.  Before we do that we want to have disseminated urls that point at that version.

peetucket:
> @peetucket but then imagine drastically changing it for v2. At least we'd have a way to get back to v1. In the work cycle we said that we didn't want to support different "with version" and "without version" behavior. I believe the intention is still to convert all non-versioned objects to objects with one user version. Before we do that we want to have disseminated urls that point at that version.

I guess I still don't see what value we get from showing that panel at all.  Even if there are multiple previous versions of an older object, it will only ever show one version, with the link to v1 (same as the current purl) and a date that could be wrong (but which we could suppress).  If and when we do migrate these older objects, the panel would then show again, with proper links to all previous versions, as it does for new objects.

I'd like @andrewjbtw and @amyehodge to make the call here.  For the context of this work, this applies to objects where there is only a single user facing version shown because it is an older object that does have any knowledge of the new "user version" distinctions.  The options are:

1. Show the "Versions" section on the PURL page, but hide the date (since it could be wrong and currently reflects the last time the PURL was published and not necessarily when the object was last accessioned).
2. Hide the "Versions" section completely.

andrewjbtw:
Returning to this, my main concern is the version date being incorrect. I checked an [item in stage that has only one version](https://argo-stage.stanford.edu/view/druid:ww621cd0795) (one user version that is also the one system version) and republishing that item does not update the date on the purl, which is as it should be. The date should only be updated if a change is made via accessioning, and that seems to be functioning correctly when an item uses the user version model. Republishing without accessioning should never update the version date.

On whether to show the panel at all when there is no user version metadata, I've come around to Peter's view that we should not show the panel on items that don't use the user version model. There's no stability to the `v1` URLs for those items, as the content could be changed completely without ever changing the "version 1" presented on the purl, and showing a `v1` link implies some level of stability. 

It's true that the even with user versions, the v1 purl could be modified, but we've implemented it in H2 in a way that requires people to create a new user version when the content changes. So while the `v1` link for a user versioned item might show updated metadata at a later date, it should not show different content. When we get around to applying the user version model to everything, we're going to have to make some tough decisions on whether to apply the same rules to Argo and/or other deposit application UIs.

amyehodge:
I concur with @andrewjbtw on this. Can we get this work completed and merged?

mjgiarlo (approved):

peetucket:
Merged.  Will go out with next dependency update release of PURL.

PR #1276, opened 2025-08-05
point manifests to version filepaths

dnoneill:
part of https://github.com/sul-dlss/sul-embed/issues/2899

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -232,9 +232,17 @@ class Iiif3PresentationManifest < IiifPresentationManifest
     img_res
   end
 
+  def stacks_version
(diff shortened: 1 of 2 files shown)


app/models/iiif3_presentation_manifest.rb
```diff
@@ -232,9 +232,13 @@ def image_resource(resource)
     img_res
   end
 
+  def stacks_version
+    controller.instance_variable_get(:@version).version_id
```
jcoyne, line 236: Can you pass this in as a parameter to the initializer?  The only reason this file needs to use the controller is for view helper methods.  We don't want to increase the coupling.

jcoyne (changes requested):

PR #1279, opened 2025-08-08
fix v3 manifest validator errors

dnoneill:
This fixes all but one error. The last error is because we are adding a pdf to the image canvas. Basically we have image01, image02, fullpdf as its own canvas. We shouldn't do this but becasue of the mixed nature of our content I would like this to be a separate issue.

Issues addressed in this PR

Before:
Screenshot 2025-08-08 at 1 23 41 PM

After:
Screenshot 2025-08-08 at 1 21 48 PM

### DIFF
--- a/app/models/iiif3_presentation_manifest.rb
+++ b/app/models/iiif3_presentation_manifest.rb
@@ -30,8 +30,8 @@ class Iiif3PresentationManifest < IiifPresentationManifest
 
     manifest = iiif_manifest_class.new(manifest_data)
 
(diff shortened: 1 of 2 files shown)


dnoneill:
I thought the renaming of service to services might break https://github.com/sul-dlss/sul-embed/blob/b5180b8728055681710e6ac5de1623c372064d73/app/javascript/controllers/file_auth_controller.js#L192 but from what I can tell it doesn't.

PR #1283, opened 2025-08-11
Bring methods from mods_display gem into SubjectComponent

### LINKED ISSUE #1274: Move link_to_mods_subject and link_mods_subjects locally
Currently they are in mods_display, but they are not used anywhere else.

marlo-longley:
Closes #1274

### DIFF
--- a/app/components/subject_component.html.erb
+++ b/app/components/subject_component.html.erb
@@ -3,12 +3,12 @@
   <%= render TableComponent.new(label_id:) do %>
     <tbody>
       <% document.mods.subject.each do |field| %>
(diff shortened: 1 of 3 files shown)


jcoyne (changes requested):
Can you bring the specs along too? https://github.com/sul-dlss/mods_display/blob/5583ac5a353ec32316871c4ceb27712f5bd246d1/spec/helpers/record_helper_spec.rb#L261

jcoyne:
I think the missing field names is orthogonal to this change. The missing fields are same both before and after this one.

marlo-longley:
@jcoyne thanks for your help -- added tests.

jcoyne (approved):

Issue #1285, opened 2025-08-13
Update fixture metadata from production


jcoyne:
The fixture metadata in spec/fixtures/document_cache is legacy metadata.  It should be updated from prod:

e.g.:
```
mkdir -p spec/fixtures/document_cache/bc/854/fy/5899/bc854fy5899/versions
scp -r lyberadmin@purl-fetcher-prod:/stacks/bc/854/fy/5899/bc854fy5899/versions spec/fixtures/document_cache/bc/854/fy/5899/bc854fy5899/
```

Then run tests and resolve any errors.

PR #1288, opened 2025-08-13
Add a script for keeping the metadata up to date

jcoyne:
Both the file structure and content
Ref #1285


jcoyne:
Yes, the filestructure is part of what it means to be up to date.  This is what the files currently look like in /stacks.

dnoneill:
could the commit be updated to `Add a script for keeping the metadata and filestructure up to date`? I think it might be faster to understand in the history. Otherwise this looks great.

dnoneill (approved):

PR #1303, opened 2025-08-14
Look for meta.json in the version directory

jcoyne:
If it's not present, look at the legacy location for meta.json
Fixes #1500

### DIFF
--- a/app/services/resource_retriever.rb
+++ b/app/services/resource_retriever.rb
@@ -78,7 +78,7 @@ class ResourceRetriever
   end
 
   def meta_json_path
-    Settings.purl_resource.meta
+    Settings.purl_resource.versioned.meta
   end
 
   def cocina_path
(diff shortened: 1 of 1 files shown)


app/services/resource_retriever.rb
```diff
@@ -78,7 +78,7 @@ def public_xml_path
   end
 
   def meta_json_path
-    Settings.purl_resource.meta
+    Settings.purl_resource.versioned.meta
```
dnoneill, line 81: Don't we need to update this in settings for this to do what we want? https://github.com/sul-dlss/purl/blob/main/config/settings.yml#L27

dnoneill: Oh I see it is the difference between druid.meta_json and druid/meta.json. I think there might be a typo or something. I don't think meta.json is a path. I have tried it with multiple druids and it doesn't work and we have migrated everything.

dnoneill (approved):

PR #1310, opened 2025-08-15
Use real fixtures for testing

### LINKED ISSUE #1296: Stop using fg019pm1396 and wr231qr2829 in tests
They no longer exist in argo.

jcoyne:
Fixes #1296


jcoyne:
@dnoneill I don't understand why this would influence that decision.  This is only test changes.

dnoneill (approved):

PR #1312, opened 2025-08-15
v3 manifest metadata pppulate with all subjects including subjects in…

dnoneill:
… structuredValue

### DIFF
--- a/app/models/iiif3_metadata_writer.rb
+++ b/app/models/iiif3_metadata_writer.rb
@@ -121,7 +121,7 @@ class Iiif3MetadataWriter
 
   def subjects
     vals = Array(cocina_descriptive['subject']).filter_map do
-      it['value'] if it['type'] == 'topic'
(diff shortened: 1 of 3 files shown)


app/models/iiif3_metadata_writer.rb
```diff
@@ -168,7 +168,14 @@ def iiif_key_value(label, values)
     { 'label' => { en: [label] }, 'value' => { en: values.compact_blank } }
   end
 
-  def structured_values(field)
-    field['structuredValue'].presence ? field['structuredValue'].pluck('value') : [field['value']]
+  def structured_values(field, field_key = 'value')
+    # zf119tw4418 has a structuredValue which has 4 items, the first one is a structuredValue
+    if field['structuredValue'].presence
+      field['structuredValue'].flat_map do |r|
```
jcoyne, line 174: Can we use a more descriptive name for this variable instead of `r`?

PR #1321, opened 2025-08-19
Reorganize iiif2 manifest spec to use context blocks

jcoyne:
Also updated bb001dq8600 from prod


spec/requests/iiif2_manifest_spec.rb
```diff
@@ -3,209 +3,228 @@
...
-    expect(rendering).to include(
-      '@id' => 'https://stacks.stanford.edu/file/zf119tw4418/zf119tw4418_31_0000.pdf',
-      'label' => 'Download Object 1',
-      'format' => 'application/pdf'
-    )
-  end
+      expect(json).not_to include('service')
+    end
 
-  it 'does not advertise downloadable files for stanford-only no-download content' do
```
dnoneill, line 205: Why is this being deleted?

PR #1330, opened 2025-08-19
Add more documentation about fixtures

jcoyne:


dnoneill:
I think this is fine but it might be nice if the commit message indicated it is just increased documentation about the collections, not all druids.

PR #1339, opened 2025-08-20
Simplify subject component

jcoyne:
Subjects with names do not have roles. Genres do not have names.

### DIFF
--- a/app/components/subject_component.rb
+++ b/app/components/subject_component.rb
@@ -15,17 +15,10 @@ class SubjectComponent < ViewComponent::Base
     subject.present? || genre.present?
   end
 
-  def link_mods_subjects(subjects)
-    subjects.map { link_to_mods_subject(it) }
-  end
-
-  # @param [ModsDisplay::Name::Person, String] subject
-  def link_to_mods_subject(subject)
-    subject_text = subject.respond_to?(:name) ? subject.name : subject
-    if subject.respond_to?(:roles) && subject.roles.present?
-      "#{subject_text} (#{subject.roles.join(', ')})"
-    else
-      subject_text
+  # @param [Array<ModsDisplay::Name::Person, String>] subjects
+  def expand_subject_name(subjects)
+    subjects.map do |subject|
+      subject.respond_to?(:name) ? subject.name : subject
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/components/subject_component.rb
```diff
@@ -15,17 +15,10 @@ def render?
...
   end
 
-  def link_mods_subjects(subjects)
-    subjects.map { link_to_mods_subject(it) }
-  end
-
-  # @param [ModsDisplay::Name::Person, String] subject
-  def link_to_mods_subject(subject)
-    subject_text = subject.respond_to?(:name) ? subject.name : subject
-    if subject.respond_to?(:roles) && subject.roles.present?
```
dnoneill, line 25: I don't think we want to get rid of this?

jcoyne: Why?  Is there such a thing as a subject that has a role?

dnoneill: https://github.com/sul-dlss/cocina-models/blob/70b258e920645dbfb20b8925db62351ab1b042d7/spec/cocina/models/validators/description_types_validator_spec.rb#L350-L389

dnoneill: Sorry I was looking in the note field. It may be old https://github.com/sul-dlss/cocina-models/blob/70b258e920645dbfb20b8925db62351ab1b042d7/spec/cocina/models/mapping/descriptive/mods/subject_name_spec.rb#L922-L931 has it but it isn't on main. I think we should track down if this assumption is correct before removing it.

jcoyne: That is so weird. I've asked Arcadia on the infrastructure Slack to weigh in.

jcoyne: She fixed the metadata on that item, but we're still not hitting that code on purl: https://purl.stanford.edu/dv559gn8984

jcoyne: ModsDisplay::Name::Person does not have a `roles` method. https://github.com/sul-dlss/mods_display/blob/5583ac5a353ec32316871c4ceb27712f5bd246d1/lib/mods_display/fields/name.rb#L104. This is just dead code.

PR #1343, opened 2025-08-20
Consolidate book fixtures

jcoyne:
We don't want to manage more metadata than we need to test.


dnoneill:
@jcoyne three of manifest specs are failing

jcoyne:
https://github.com/sul-dlss/purl/pull/1368 will remove one of the specs that is failing.

jcoyne:
Seems like we do need this fixture as there are no others that are world access + OCR.

PR #1345, opened 2025-08-20
Avoid using a whole fixture, when we can just use the relevant metada…

jcoyne:
…ta for testing

This was not a great fixture to use as it's an out of date style of book because it has image constituents and doesn't shelve the OCR.


dnoneill:
Why would we do this if we are keeping the fixture? Isn't this just making the file longer for information we already have?

PR #1347, opened 2025-08-21
Fix example wording for spec

jcoyne:


marlo-longley (approved):

marlo-longley:
@jcoyne needs a rebase now

PR #1368, opened 2025-08-21
Use copyright from Cocina data

### LINKED ISSUE #1354: Get Copyright (in requiredStatement) from access.copyright in the Cocina

jcoyne:
Fixes #1354


spec/requests/iiif3_manifest_spec.rb
```diff
@@ -145,12 +145,6 @@
     end
   end
 
-  it 'properly decodes XML entities into their UTF-8 characters' do
```
dnoneill, line 148: Same as above

jcoyne: Same as above 😉

dnoneill (approved):

PR #1407, opened 2025-08-25
update hx163dc5225

### LINKED ISSUE #1316: Update or replace hx163dc5225 as a fixture
Currently hx163dc5225 is a fixture, but the tests break if we update it from production.

dnoneill:
closes #1316


dnoneill:
oops. yes.

PR #1411, opened 2025-08-25
only add renderings that can be downloaded

dnoneill:
ref https://github.com/sul-dlss/sul-embed/issues/2901#issuecomment-3214777738


jcoyne (commented):
Can you add a test to this one please?

spec/requests/iiif3_manifest_spec.rb
```diff
@@ -71,6 +71,28 @@
     end
   end
 
+  context 'with location specific, no download files' do
```
jcoyne, line 74: Should we put this in the Iiif3PresentationManifest spec instead? It seems this is mostly about the generated manifest and not the http request.

jcoyne: That does seem like a good idea. Should we make a follow up issue?

jcoyne (approved):

PR #1424, opened 2025-08-26
Decouple authors from xml

### LINKED ISSUE #1400: Refactor PurlVersion#authors to use cocina rather than public xml

jcoyne:
Fixes #1400


app/models/purl_version.rb
```diff
@@ -263,7 +263,7 @@ def publication_date
     end
 
     def authors
-      @authors ||= ::Metadata::Authors.call(mods_ng_document)
+      cocina_display.contributors.map(&:display_name)
```
dnoneill, line 266: I know this is probably tested already in the context of the cocina gem but it might be nice to have a test to make sure it is working properly in the context of purl. It would also be a pretty good debugging tool if something goes wrong in the gem.

PR #1426, opened 2025-08-27
Create an Altmetrics component

jcoyne:


dnoneill:
Can you remove these comments. Seem like they aren't necessary for the component but we don't want to keep them around https://github.com/sul-dlss/purl/blob/a5bbafc3757dad20261e4e43c567d221973eb87e/app/models/purl_version.rb#L244

dnoneill:
@jcoyne Is there a reason it would be better in altmetrics. I think delegating to purl_version is a better solution than moving it there.

jcoyne:
@dnoneill maybe you can explain your reasoning for that?  My though is that a method that is only used by the AltmetricComponent should only exist in the AltmetricComponent.

dnoneill:
For some reason I thought it was used elsewhere, something very similar is in app/models/iiif3_metadata_writer.rb which caused some confusion.

dnoneill (approved):

PR #1437, opened 2025-08-28
Remove remote data fetching

jcoyne:
This is not something we use, but it adds unnecessary complexity

### DIFF
--- a/app/services/resource_retriever.rb
+++ b/app/services/resource_retriever.rb
@@ -39,7 +39,7 @@ class ResourceRetriever
 
   def version_manifest_resource
     @version_manifest_resource ||= cache_resource(:version_manifest) do
-      fetch_resource(:version_manifest, Settings.stacks.version_manifest_path)
+      fetch_resource(Settings.stacks.version_manifest_path)
     end
   end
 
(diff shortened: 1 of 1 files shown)


jcoyne:
@dnoneill I don't understand how.  Our development environment should be just like our production environment, which means that any data you want to display you must have locally.  This is how all other projects work. We should not have separate code that only runs in development, otherwise how can you trust the code to work the same on prod?

jcoyne:
@dnoneill All we have to do is use the fixture harvester.  I think it's really bad to make this work differently in prod and development.

dnoneill:
Fixture harvester only works for prod and it goes through everything. I think it would be good to have a plan to make this not arduous on the developers. I agree consistency is good but I don't think we should necessarily do that at the detriment of the developers.

jcoyne:
@dnoneill I think that stuff is removed in https://github.com/sul-dlss/purl/pull/1410

dnoneill:
Okay I changed the paths and I am still getting errors for files I have locally. It is possible this doesn't work without https://github.com/sul-dlss/purl/pull/1410. Though those paths are only changed in test.yml. This is a tangle that will have to wait until Tuesday.

PR #1453, opened 2025-09-02
Use cocina_display for publication_date

jcoyne:


spec/components/altmetrics_component_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+                            }
+          }
+        JSON
+        allow(Honeybadger).to receive(:notify)
+
+        render_inline(component)
+      end
+
+      it 'returns nil' do
+        # expect(instance.publication_date).to be_nil
```
marlo-longley, line 79: Did you want to include this expectation?

PR #1460, opened 2025-09-03
Remove unnecessary conditional

jcoyne:
Mods is always present in the public metadata


marlo-longley:
@jcoyne looks good but needs rebase

PR #1461, opened 2025-09-03
Extract components from the show page

jcoyne:
This will make it easier to test the show page without having to do feature tests for each case. Feature tests are much slower than component tests

### DIFF
--- a/app/components/show/body_component.html.erb
+++ b/app/components/show/body_component.html.erb
@@ -1,0 +1,7 @@
+    <div class="record-sections col-lg-8">
+      <%= render "mods_abstract_contents", mods: %>
+      <%= render ModsDescriptionComponent.new(mods:) %>
+      <%= render ModsContributorsComponent.new(mods:) %>
+      <%= render SubjectComponent.new(mods:) %>
+      <%= render BibliographicComponent.new(mods:) %>
+    </div>
--- a/app/components/show/body_component.rb
+++ b/app/components/show/body_component.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Show
+  class BodyComponent < ViewComponent::Base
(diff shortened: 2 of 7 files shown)


app/components/show/body_component.html.erb
```diff
@@ -0,0 +1,9 @@
+    <div class="record-sections col-lg-8">
+      <% if version.mods? %>
```
marlo-longley, line 2: Didn't you delete this method here? https://github.com/sul-dlss/purl/pull/1460/files

app/components/show/body_component.html.erb
```diff
@@ -0,0 +1,9 @@
...
+        <%= render "mods_abstract_contents", document: version %>
+        <%= render ModsDescriptionComponent.new(document: version) %>
```
marlo-longley, line 4: I am finding the mixing of `version` and `document` across the app a bit confusing. Should this be named `document` in the body_component rather than `version`? Take [the FindItComponent for example](https://github.com/sul-dlss/purl/blob/a0040ac147cc90499479653538109d0ca1196952/app/components/find_it_component.rb#L4) -- it takes both `version` and `document` as distinct arguments, implying that these words mean different things.

app/components/show/body_component.html.erb
jcoyne: Yes, but that hasn't been merged yet.  One pr will need to be rebased depending on which merges first.

app/components/show/body_component.html.erb
jcoyne: It is confusing, but this is as it was in https://github.com/sul-dlss/purl/pull/1461/files#diff-53665237255233d83338a25e74997961004f14a8d124aa541a8bab9155c2f52cL58

PR #1463, opened 2025-09-03
Extract a class to model the meta.json data for releases

jcoyne:


app/models/purl_resource.rb
```diff
@@ -36,17 +36,8 @@ def version(version_id)
...
-  end
-
-  def released_to_searchworks?
-    meta_json.fetch('searchworks')
-  end
-
-  def released_to_earthworks?
-    meta_json.fetch('earthworks')
+  def releases
+    @releases ||= Releases.new(meta_json)
```
dnoneill, line 40: I always get confused about Memoization but could this cause a problem if the user switches between versions and one is released to searchworks and the other isn't? Or maybe that isn't an issue because it should just be most recent.

jcoyne: It doesn't matter. You don't release versions, you only release a whole object.

dnoneill (approved):

PR #1469, opened 2025-09-03
Remove the ability to preview mods

jcoyne:
As we migrate to Cocina for metadata display, this will no longer work


dnoneill:
What did this do? I see there is a preview route but it didn't take a druid so it just show the mods schema? What was the intention?

jcoyne:
@dnoneill I wasn't here for this, but I think the idea was to paste in mods and it would draw the show page for whatever you had provided.  I don't think this has been used since the repository swtiched to use cocina as the canonical format.

dnoneill (approved):

PR #1472, opened 2025-09-03
Decouple iiif manifest building from xml data

### LINKED ISSUE #1403: Decouple iiif manifest building from content_metadata
Use the cocina structural instead.

dnoneill:
closes #1403


app/models/structural_metadata.rb
```diff
@@ -20,8 +20,28 @@ def resources
...
+
+  def file_sets
+    @file_sets ||= Array(json['contains']).map { FileSet.new(json: it, druid:) }
+  end
+
+  def local_files
+    @local_files ||= file_sets.flat_map(&:files)
+  end
+
   def find_file_by_filename(filename)
```
dnoneill, line 43: This is used by external_file function which it doesn't seem like we are using anywhere. I am confused by what external_file is even doing. That would be helpful

dnoneill: See comment below

app/models/iiif3_presentation_manifest.rb
```diff
@@ -80,31 +81,64 @@ def provider
...
+  end
+
+  # Cocina only lists druids for collection objects, not filesets
+  # We need to go grab the PurlResource for the druid
+  # It seems like we are only doing this for type images. I am not sure why
+  def add_virtual_object_canvases(manifest)
+    # For each valid virtual object image, create a canvas for its thumbnail
+    structural_metadata.members&.each do |member_druid|
+      purl_version = PurlResource.find(member_druid.delete_prefix('druid:')).version(:head)
+      # We are using .thumbail here to get the first image in the object
```
jcoyne, line 124: We called this `thumbnail` when we were using this for other thumbnail stuff, but now, I think something like "first_image_fileset" would be much clearer.

app/models/structural_metadata/file_set.rb
```diff
@@ -13,8 +13,77 @@ def type
       json['type']
     end
 
+    def id
+      json['externalIdentifier'].chomp('/')
```
jcoyne, line 17: Why do we `chomp('/') here?

app/models/structural_metadata/file.rb
```diff
@@ -33,8 +35,32 @@ def width
...
+
     def access
       @access ||= FileAccess.new json['access']
     end
+
+    def role
+      json['use']
+    end
+
+    def image_file?
```
jcoyne, line 62: Can we merge this with `image?`

app/models/structural_metadata/file_set.rb
```diff
@@ -13,8 +13,77 @@ def type
...
+    end
+
+    def label
+      json['label']
+    end
+
+    def cocina_id
```
jcoyne, line 24: Can you add a comment about what this method is doing and why we need it?

app/models/structural_metadata/file_set.rb
```diff
@@ -13,8 +13,77 @@ def type
...
+      files - [primary, thumbnail_canvas].compact - supplementing_resources
+    end
+
+    def supplementing_resources
+      return [] if media_file.blank?
+
+      files.select { |file| file.mimetype == 'text/vtt' }
+    end
+
+    def thumbnail_canvas
```
jcoyne, line 83: This seems like a weird name, because "canvas" is a IIIF term, but we don't want any IIIF terms in the metadata model.  Can it be `thumbnail_file_for_media`?

spec/model/iiif_presentation_manifest_spec.rb
```diff
@@ -146,7 +146,7 @@
         rendering = json['sequences'].first['rendering']
         expect(rendering).to include(
           '@id' => 'https://stacks.stanford.edu/file/zf119tw4418/zf119tw4418_31_0000.pdf',
-          'label' => 'Download Full PDF',
+          'label' => 'Download zf119tw4418_31_0000.pdf',
```
jcoyne, line 149: Why is this changing?

jcoyne (changes requested):
This looks really good. I just have a few small questions and a few spots where I think this can be improved.

app/models/iiif_presentation_manifest.rb
```diff
@@ -303,13 +301,13 @@ def thumbnail_resource
 
     thumb
   end
-  # rubocop:enable Metrics/AbcSize
+  # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/AbcSize
 
-  def rendering_resource(resource, label: "Download #{resource.label}", profile: nil)
+  def rendering_file(file, label: "Download #{file.label}", profile: nil)
```
jcoyne, line 306: Should this label be `file.fileset_label`?

spec/model/iiif_presentation_manifest_spec.rb
dnoneill: Yeah I don't think that should be changed. I will look at that

app/models/iiif3_presentation_manifest.rb
dnoneill: I agree. I kept confusing myself when fixing this up because it originally returned a file and we needed the fileset so I think that makes a lot of sense.

app/models/iiif_presentation_manifest.rb
```diff
@@ -128,37 +113,49 @@ def body
...
+    # For each local file image, create a canvas
+    page_image_files.each do |file|
+      sequence.canvases << canvas_for_file(file)
+    end
+
+    # For each valid virtual object image, create a canvas for its thumbnail
+    structural_metadata.members&.each do |member_druid|
+      purl_version = PurlResource.find(member_druid.delete_prefix('druid:')).version(:head)
+      # We are using .thumbail here to get the first image in the object
+      thumbnail_file = purl_version.structural_metadata.thumbnail.image_file
```
jcoyne, line 131:
```suggestion
      thumbnail_file = purl_version.thumbnail
```

app/models/iiif3_presentation_manifest.rb
jcoyne: Can you make this name change?

app/models/iiif_presentation_manifest.rb
dnoneill: Yeah this is going to work for v2 but not v3. I am still trying to figure out how to untangle this.

PR #1475, opened 2025-09-04
Move tests for metadata fields to compnent tests

jcoyne:
The feature tests run slow, so a component test is more appropriate for these


dnoneill:
typo in commit message for component. it is component not compnent

PR #1484, opened 2025-09-05
Refactor test to show more of what we are expecting, rather than samp…

jcoyne:
…ling individual fields


marlo-longley:
@jcoyne this looks good but needs a rebase

marlo-longley (approved):

PR #1486, opened 2025-09-05
Simplify the test for virtual objects

jcoyne:
Only the parent object is interesting, the child objects are just regular images


dnoneill:
rebase

marlo-longley (approved):

PR #1490, opened 2025-09-05
Replace mods contact with cocina contact

jcoyne:


app/components/contact_component.rb
```diff
@@ -0,0 +1,19 @@
...
+  def initialize(version:)
+    @version = version
+    super()
+  end
+
+  attr_reader :version
+
+  # TODO: https://github.com/sul-dlss/cocina_display/issues/97
+  def contacts
+    @contacts ||= Array(version.cocina.dig('description', 'access', 'accessContact')).select { it['type'] == 'email' }
```
dnoneill, line 13: This is in cocina_display as contact_information

corylown: Soon to be `contact_email_display_data`

dnoneill:
@jcoyne I forgot to link to the PR. https://github.com/sul-dlss/cocina_display/pull/129. Cory is going to be updating the name but it is in there.

PR #1495, opened 2025-09-05
Generate annotation ids using the fileset rather than a file

jcoyne:


app/models/iiif3_presentation_manifest.rb
```diff
@@ -236,22 +236,22 @@ def image_rendering_for_fileset(fileset)
...
     anno_page = IIIF::V3::Presentation::AnnotationPage.new
-    anno_page['id'] = annotation_page_url(resource_id: resource.id)
-    anno_page.items << annotation_for_resource(resource)
+    anno_page['id'] = annotation_page_url(resource_id: file.fileset_id)
+    anno_page.items << annotation_for_resource(file)
     anno_page
   end
 
-  def annotation_for_resource(resource)
+  def annotation_for_resource(file)
```
dnoneill, line 246: can this be changed annotation_for_file

PR #1496, opened 2025-09-05
add twitter title and image

### LINKED ISSUE #1425: No cards show in Twitter
Paste a Purl url into the Tweet Composer.

No card is loaded.

Possibly because the required field twitter:title is missing.

dnoneill:
closes #1425

### DIFF
--- a/app/views/purl/show.html.erb
+++ b/app/views/purl/show.html.erb
@@ -27,12 +27,14 @@
     graph.title @version.display_title
     graph.description @version.description.truncate(200) if @version.description
     graph.type 'og:website'
-    graph.image purl_url(@purl, format: :jpg)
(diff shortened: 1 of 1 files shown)


app/views/purl/show.html.erb
```diff
@@ -33,6 +33,8 @@
 
   twitter_card('summary_large_image') do |card|
     card.site Settings.twitter.site
+    card.title @version.display_title
+    card.image purl_url(@purl, format: :jpg)
```
jcoyne, line 37: Isn't this just a redirect? Shouldn't we give the thumbnail?

jcoyne: It would be better for us if we can just put the image path here, then we only have to handle 1 request instead of two.

dnoneill: Done

PR #1497, opened 2025-09-05
Remove unnecessary method

jcoyne:
We can do the same with exception handling


app/services/resource_retriever.rb
```diff
@@ -58,11 +38,11 @@ def version_manifest_path
   end
 
   def meta_json_resource
-    @meta_json_resource ||= resource_cache.get(meta_json_path, cache_key(:meta))
+    @meta_json_resource ||= File.open(meta_json_path)
```
dnoneill, line 41: Shouldn't the .read go on here? I think it is a little weird to have it on the variable whose name indicates it is the thing no a File.open

dnoneill: Then why aren't we using meta_json_body here?

jcoyne: You may be confused about the order.  `meta_json_body` calls `meta_json_resource`.

PR #1509, opened 2025-09-09
fix thumbnail check, update height / width to be accurate

### LINKED ISSUE #1508: Citation only items in Mirador show "1 of 0" and don't show an image thumbnail
Citation-only items in Mirador are not showing a thumbnail. Instead, the viewer is blank and the page number indicator says "1 of 0".

Examples:

https://purl.stanford.edu/fw748fx1968
https://purl.stanford.edu/xq083mf5798

Image

dnoneill:
closes #1508

### DIFF
--- a/app/models/structural_metadata/file.rb
+++ b/app/models/structural_metadata/file.rb
@@ -27,12 +27,42 @@ class StructuralMetadata
       "#{Settings.stacks.url}/image/iiif/#{druid}%2F#{ERB::Util.url_encode(::File.basename(filename, '.*'))}"
     end
 
-    def height
+    def image_height
       json.dig('presentation', 'height')
     end
 
-    def width
+    def image_width
       json.dig('presentation', 'width')
+    end
+
+    def height
+      return image_height if viewable?
+
+      thumbnail_height
+    end
+
+    def width
+      return image_width if viewable?
+
+      thumbnail_width
+    end
+
+    # thumbnail width/height sets long edge to 400px
+    # then calculates the correct dimensions for the short edge to preserve aspect ratio
+    def thumbnail_height
+      if height >= width
+        400
+      else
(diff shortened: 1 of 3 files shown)


app/models/structural_metadata/file.rb
```diff
@@ -55,6 +83,10 @@ def access
       @access ||= FileAccess.new json['access']
     end
 
+    def full_access?
```
jcoyne, line 86: This method name doesn't mean anything to me.   How about, `viewable_by_any_auth?` instead?

app/models/structural_metadata/file.rb
```diff
@@ -35,6 +35,34 @@ def width
...
+      thumbnail_height
+    end
+
+    def manifest_width
+      return width if full_access?
+
+      thumbnail_width
+    end
+
+    def thumbnail_height
```
jcoyne, line 50: Can you add documentation about what you are doing here?  It seems to be preserving the aspect ratio and ensuring the long edge is 400px?

jcoyne (changes requested):

app/models/structural_metadata/file.rb
dnoneill: Maybe viewable?

app/models/structural_metadata/file.rb
```diff
@@ -35,6 +35,34 @@ def width
       json.dig('presentation', 'width')
     end
 
+    def manifest_height
```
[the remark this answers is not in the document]
dnoneill: I think it should go in the file just be renamed. I think we want to keep it this way. The viewable width/height for things like citation only are always going to be the thumbnail size not the actual width/height.

jcoyne (approved):

PR #1519, opened 2025-09-10
Add contributors from Cocina

jcoyne:


app/components/contributor_component.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  attr_reader :contributor
+
+  def orcid
+    @orcid ||= contributor.identifiers.find { it.type == 'ORCID' }
+  end
+
+  def name
+    contributor.display_name(with_date: true)
```
dnoneill, line 16: cruft?

app/components/contributors_component.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= render SectionComponent.new(label: 'Creators/Contributors', label_id:) do %>
+  <%= render TableComponent.new(label_id:) do %>
+    <% contributors_by_role.each do |label, contributors| %>
+      <%= render TableBodyComponent.new(label:) do |component| %>
+        <% contributors.each do |contributor| %>
+          <%= component.with_value do %>
+            <%= render ContributorComponent.new(contributor:) %>
```
dnoneill, line 7: Does this need to be a separate component? If it does can it be named slightly differently. I had to take a beat to figure out the difference between contributors and contributor. The contributor component is only 4 lines of html. Also if we are going to use a componet wouldn't it make sense to use with_collection instead of using a loop.

dnoneill (commented):
Visually this looks good.

PR #1528, opened 2025-09-12
Use form_display_data to get resource type from cocina

jcoyne:


config/locales/en.yml
```diff
@@ -23,3 +23,7 @@ en:
   date:
     formats:
       short: "%b %-d, %Y"
+  cocina_display:
+    field_label:
+      form:
+        resource_type: Resource Type
```
thatbudakguy, line 29: Why override this here? Let's just change it in cocina_display.

PR #1537, opened 2025-09-15
add record with parallelValue in title and contributor

dnoneill:
I considered just adding the cocina into the iiif3 metadata writer test but I am pretty sure we are going to want this record for when we switch title/contributors over to cocina_display


config/settings/development.yml
```diff
@@ -5,6 +5,7 @@ stacks:
 landing_page_druids:
   - bb000qr5025 # image where jp2 is not downloadable
   - bb001dq8600 # image, stanford-only
+  - bb070yy8209 # book, with transliterated fields
```
jcoyne, line 8: 🥇

spec/model/iiif3_metadata_writer_spec.rb
```diff
@@ -551,5 +551,15 @@ def field(field_name)
         expect(field_value('Identifier')).to eq ['LCCN: 30003962', 'localnumber', 'nosourceident']
       end
     end
+
+    context 'when the title has parallelValues' do
+      let(:cocina_descriptive) do
+        PurlResource.find('bb070yy8209').version(:head).cocina['description']
+      end
+
+      it 'extracts all the identifiers correctly' do
```
jcoyne, line 560:
```suggestion
      it 'extracts all the titles correctly' do
```

jcoyne (changes requested):

Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/sul-dlss/cocina_display.git
-  revision: 119395f29aac44727932db604a61d8a5b6cb7a49
+  revision: 185c022b1bb3bda8648be60cca222d0c46b8da2c
```
jcoyne, line 3: Is this change required in this PR?

jcoyne (changes requested):

Gemfile.lock
dnoneill: I think that was to get it to the latest version which at the time was necessary. I don't think it is if we are on the most recent version now

jcoyne (approved):

PR #1568, opened 2025-09-22
Use description and keywords from cocina rather than MODS

### LINKED ISSUE #1565: Get keywords and description from Cocina rather than mods
In HeadMetadataComponent

jcoyne:
Fixes #1565


Gemfile
```diff
@@ -17,6 +17,7 @@ gem 'okcomputer' # application monitoring
 gem "propshaft"
 gem "puma", "~> 7.0" # web server for development
 gem "rails", "~> 8.0.0"
+gem "rails_autolink", "~> 1.1"
```
marlo-longley, line 20: I am wondering if there's a way we can migrate off the old auto_link (which seems only to be used 3 times in the app) without introducing a new library. Would the rails LinkSanitizer or similar work? I may be missing something about what auto_link provided.

dnoneill (approved):

PR #1575, opened 2025-09-23
Add herb linter

jcoyne:


marlo-longley:
Do we want to include any info on running this locally? I ran `npx @herb-tools/linter` but not sure if we want that in README. People will need to fix up their PRs locally.

PR #1598, opened 2025-09-24
Add coverage tooling

jcoyne:


spec/spec_helper.rb
```diff
@@ -1,7 +1,22 @@
...
+
+  # Track coverage for all source files
+  add_group 'Components', 'app/components'
+  add_group 'Services', 'app/services'
+
+  # Set minimum coverage expectation
+  minimum_coverage 97
+  minimum_coverage_by_file 82
+
+  # Refuse to merge results unless they're from the same commit
```
marlo-longley, line 17: I don't understand how this enforces the same commit. I get that it will prevent merged results older than 1 hour.

marlo-longley (approved):

PR #1601, opened 2025-09-24
Remove unused link

jcoyne:
This hasn't been used since we switched to stimulus and it is just a link to the current page that we are on.


jcoyne:
@marlo-longley rebased.

marlo-longley (approved):

PR #1602, opened 2025-09-24
Move withdrawn_cocina to PurlVersion

jcoyne:


app/controllers/purl_controller.rb
```diff
@@ -37,7 +37,7 @@ def show
 
       format.json do
         if @version.withdrawn?
-          render json: withdrawn_cocina
+          render json: @version.withdrawn_cocina
```
dnoneill, line 40: I think this needs .to_json?

jcoyne: .to_json is automatically called.

jcoyne: Yes, it's built into rails.

dnoneill (approved):

PR #1603, opened 2025-09-24
Extract ItemType class from PurlVersion

jcoyne:
This groups all the methods about the item's type into a single class that has no other responsibilities.

### DIFF
--- a/app/models/purl_version.rb
+++ b/app/models/purl_version.rb
@@ -68,6 +68,7 @@ class PurlVersion # rubocop:disable Metrics/ClassLength
   end
 
   delegate :containing_collections, to: :structural_metadata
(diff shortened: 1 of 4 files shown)


app/models/purl_version.rb
```diff
@@ -103,22 +104,18 @@ def iiif2_manifest?
     return false if collection?
 
     resource_types = structural_metadata.file_sets.map(&:type)
-    if (image? || book? || map?) &&
+    if (item_type.image? || item_type.book? || item_type.map?) &&
```
marlo-longley, line 107: I am just curious why you put some of the methods into delegation like `collection?` but left others like `item_type.book?`

PR #1605, opened 2025-09-24
Use rails internal routing mechanisms where available

jcoyne:


app/models/purl_resource.rb
```diff
@@ -14,6 +14,11 @@ def self.find(id)
     PurlResource.new(id:)
   end
 
+  # Enables us to use purl_url route helper
```
dnoneill, line 17: Are you sure this is necessary? I removed it and things still seemed to work.

dnoneill (changes requested):
There is something wrong with this branch. Going to http://localhost:3000/zb733jx3137/version/2 gives a blank rect

app/models/purl_resource.rb
jcoyne: Yeah, you are right. Originally I had `link_to @purl` which required it, but now it doesn't.

dnoneill (approved):

PR #1612, opened 2025-09-24
Add content negotiation to iiif endpoint

### LINKED ISSUE #1609: Add IIIF content negotiation
See https://iiif.io/api/cookbook/recipe/0057-publishing-v2-and-v3/

jcoyne:
Fixes #1609

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -5,16 +5,14 @@ class IiifController < ApplicationController
   before_action :load_version
 
   def manifest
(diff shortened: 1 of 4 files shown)


marlo-longley:
`iiif_version` is set in route defaults in several places. With this approach, can those be deleted?
See https://github.com/sul-dlss/purl/blob/2cf0f24e44fe5de5d36bca2a112dde54893ffc5f/config/routes.rb#L28

jcoyne:
@marlo-longley If they are using the v3 route, it will always give them v3.  Presumably we can retire those routes (or make them redirects), when we default to serving v3.

PR #1621, opened 2025-09-25
Remove mods endpoint

### LINKED ISSUE #1620: remove mods endpoint
As soon as https://github.com/sul-dlss/exhibits/pull/2980 is released

jcoyne:
It is no longer used by anything

Fixes #1620

No longer required since https://github.com/sul-dlss/exhibits/pull/2980 has been merged.


dnoneill (changes requested):
@jcoyne we have it in both versions of the manifest: https://purl.stanford.edu/yy816tv6021/iiif3/manifest. I would double check we can remove the endpoint. If we can we need to remove this from the manifest writer

dnoneill:
We can't get rid of it: https://stanfordlib.slack.com/archives/C056A6GDE7Q/p1759344548584359

PR #1624, opened 2025-09-25
Remove unused method

jcoyne:


dnoneill:
Method removed and replaced with https://github.com/sul-dlss/purl/commit/0a664c3d307433875b0099ac5a4a41833a874d91

PR #1630, opened 2025-09-26
Refactor embeddable_url to use current_page?

jcoyne:


app/components/show/head_metadata_component.rb
```diff
@@ -29,7 +29,7 @@ def title
     end
 
     def oembed_path(format)
-      oembed_url_template.expand(format: format, application_options: oembed_url_template_options, url: embeddable_url(druid, version_id))
+      oembed_url_template.expand(format: format, application_options: oembed_url_template_options, url: embeddable_url(version))
```
marlo-longley, line 32: Can we remove `version_id` from delegation at the top of the file? I don't think it's used otherwise in this component.

PR #1650, opened 2025-09-29
for media type only put image and media to items in iiif3 manifest

### LINKED ISSUE #922: Transcript has motivation: painting
It should be `supplementing`

dnoneill:
closes #922


app/models/iiif3_presentation_manifest.rb
```diff
@@ -105,8 +105,14 @@ def build_non_image_canvases(manifest)
       # model-viewer only supports glTF/GLB 3D models (in cocina type == 3d)
       next if three_d? && fs.type != 'https://cocina.sul.stanford.edu/models/resources/3d'
 
-      canvas = canvas_for_fileset(fs)
-      manifest.items << canvas if canvas
+      # we don't want to non-image/media items in manifest items
```
jcoyne, line 108:
```suggestion
      # we don't want to add non-image/media items in manifest items
```
???

jcoyne (changes requested):

Issue #1047, opened 2024-06-24
Abstract url not wrapping when screen resized


saseestone:
Reported via PURL feedback: https://jirasul.stanford.edu/jira/browse/PURL-1354 

Name: Hannah Frost
Email: hfrost@stanford.edu
Comment:
A link in the abstract on this page does not wrap; instead it extends into the right column, running over the text displayed there. I got a screenshot showing it.

Message sent from: https://purl.stanford.edu/vg497cb6002 
Host:
IP: 64.110.49.210
User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Safari/537.36
Viewport: width: 1197 height: 649

Here's a screen grab: 
![image](https://github.com/sul-dlss/purl/assets/14897546/3185fcba-32b8-4739-a24f-6ba7c5cbcca0)

PR #1661, opened 2025-10-01
auto_link abstract text

dnoneill:
The screenshot in https://github.com/sul-dlss/purl/issues/1047 shows that the links in abstract were hyperlinks. We aren't currently doing this.

Before:
Screenshot 2025-10-01 at 5 56 14 PM

After:

Screenshot 2025-10-02 at 10 47 39 AM


jcoyne:
@dnoneill should we be using `su-underline` here since this is a link within text?

dnoneill:
@jcoyne I am not quite sure. I am always unclear on what the parameters for using su-underline vs the default.

dnoneill:
Updated

PR #1667, opened 2025-10-03
Display related link as a block element

### LINKED ISSUE #1666: URL-only related items appearing on one line
See this item: https://purl.stanford.edu/dw947kz5195

Image

The three URLs should be on separate lines.

jcoyne:
So they don't all run together.

Fixes #1666
Before
Screenshot 2025-10-03 at 12 47 38 PM

After
Screenshot 2025-10-03 at 12 47 12 PM


jcoyne:
Yes, that's the standard amount of padding below a block element.

PR #1748, opened 2026-04-20
Use lists to render tables of contents

thatbudakguy:
This improves the visual presentation for items with tables of
contents, since we can often split them up based on delimiters.


thatbudakguy:
Example from xx402cm3448:

Screenshot 2026-04-20 at 09 11 31

jcoyne (approved):

PR #1749, opened 2026-04-21
Adds preview endpoint.

justinlittman:
This will allow Argo users to preview the rendering of PURL descriptive metadata.

### DIFF
--- a/app/views/purl/preview.html.erb
+++ b/app/views/purl/preview.html.erb
@@ -1,0 +1,8 @@
+<%= render AbstractContentsComponent.new(version: @version) %>
+<%= render DescriptionComponent.new(version: @version) %>
+<%= render ContributorsComponent.new(version: @version) %>
+<%= render SubjectComponent.new(version: @version) %>
+<%= render BibliographicComponent.new(version: @version) %>
+<%= render AccessComponent.new(version: @version) %>
+<%= render CitationComponent.new(version: @version) %>
+<%= render ContactComponent.new(version: @version) %>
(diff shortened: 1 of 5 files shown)


app/views/purl/preview.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render AbstractContentsComponent.new(version: @version) %>
+<%= render DescriptionComponent.new(version: @version) %>
+<%= render ContributorsComponent.new(version: @version) %>
+<%= render SubjectComponent.new(version: @version) %>
+<%= render BibliographicComponent.new(version: @version) %>
```
thatbudakguy, line 5: Is it dumb to suggest just rendering a single `BodyComponent` instead of these five, since that's what it does?

spec/requests/purl_preview_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'PURL preview' do
+  describe 'POST /purl_preview' do
```
thatbudakguy: is the route `/preview` or `/purl_preview`?

thatbudakguy (approved):
Fine with this as discussed, just a few small questions

app/views/purl/preview.html.erb
justinlittman: Yeah, except it is wrapped in a `<div>` with `col-lg-8`.

spec/requests/purl_preview_spec.rb
justinlittman: It was originally its own controller with this route, but I moved it into the existing controller. Fixing.

PR #1759, opened 2026-05-26
Correctly render multiple contact emails

thatbudakguy:
This ports the same logic/template used in Searchworks, which
is correct. Tested on yr306cd5969.

### DIFF
--- a/app/components/contact_component.rb
+++ b/app/components/contact_component.rb
@@ -9,9 +9,8 @@ class ContactComponent < ViewComponent::Base
   attr_reader :version
 
   delegate :cocina_display, to: :version
-  delegate :contact_email_display_data, to: :cocina_display
 
   def render?
-    contact_email_display_data.present?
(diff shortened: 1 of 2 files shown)


app/components/contact_component.rb
```diff
@@ -9,9 +9,8 @@ def initialize(version:)
   attr_reader :version
 
   delegate :cocina_display, to: :version
-  delegate :contact_email_display_data, to: :cocina_display
 
   def render?
-    contact_email_display_data.present?
+    cocina_display&.contact_email_display_data.present?
```
dnoneill, line 14: Is there a reason to get rid of the delegation? Is it doing weird things?

thatbudakguy: It seemed pointless for one method 🤷🏻 we don't do it in most of the other components or in the Searchworks version, and I wanted to copy the Searchworks one as closely as possible.

taylor-steve (approved):

